PR #87, opened 2019-05-14
fix perms, deps, build

gdestuynder:
This did not build for me "so I fixed it"

I'm not sure if that's the kind of fix you're looking for, but if yes, well, this works for me :)

In particular, I was unsure why it was rebuilding libsass, but it seems to work without at least with the newer packages. Note: this also "fixes" a bunch of vulnerable libs.

cc @mpurzynski


isaisabel:
Hi @gdestuynder,

We released a new minor version of the Navigator today which addressed the same things you mentioned here. I'm closing this pull request since the changes are now part of the current version.

PR #109, opened 2019-10-29
Fix for issue preventing default layers from being loaded

wietze:
In v2.3 of the navigator, even when the Navigator is opened without the `layerURL` parameter set, it assumes it was (due to a logic error) and will never check if the `enabled` option under `default_layers` in `config.json` is set to `true`. 

This commit fixes the issue by checking if the result of `getNamedFragmentValue()` is not null and its length greater than 0. 

Tested in Chrome 77.


isaisabel:
Hi @wietze,

Thank you for finding this bug and submitting the merge request. Everything looks good and I'll be merging it into develop, and then master, later today.

PR #131, opened 2020-02-29
Updated Dockerfile to use node user

ryanpersaud:
Per[ docker-node's best practices](https://github.com/nodejs/docker-node/blob/master/docs/BestPractices.md#non-root-user) (and general Docker best practices), the container should not run as root.  As suggested in the docker-node documentation, I added `USER node` to the end of the Dockerfile and validated that the container still runs as expected.


isaisabel:
Hi @ryanpersaud,

Thank you for submitting the pull request! We're not making changes to our master branch right now (it's frozen during the [sub-techniques beta](https://medium.com/mitre-attack/attack-subs-what-you-need-to-know-99bce414ae0b0)) but I'll merge it into one of our WIP branches so that it'll end up on master the next time we do a release.

Issue #141, opened 2020-04-08
Add "show all/hide all" sub-techniques buttons


isaisabel:
As a user of the ATT&CK Navigator, I want to be able to show or hide all sub-techniques in the layer with a single button press.

Essentially the [showSubtechniques](https://github.com/mitre-attack/attack-navigator/blob/develop/nav-app/src/app/viewmodels.service.ts#L1230) field of each technique needs to be set to true or false depending on which button was pressed.

isaisabel:
Merged into the subtechniques-3.1 branch.

PR #146, opened 2020-04-13
Add toggle to show/hide all sub-techniques

clemiller:
Added toggle button that expands or condense all sub-techniques in the layer.

Resolves issue #141


nav-app/src/app/viewmodels.service.ts
```diff
@@ -369,6 +369,7 @@ export class ViewModel {
 
 
     hideDisabled: boolean = false; //are disabled techniques hidden?
+    showAllSubtechniques: boolean = false;
```
isaisabel, line 372: Setting this boolean with the control doesn't actually _do_ anything, and it isn't the expected functionality of the control. 
- Clicking the button should show/hide all sub-techniques, but this boolean isn't referenced anywhere from the Matrix layouts to satisfy that requirement. It's only ever referenced to set the value when the user clicks the toggle button.
- There probably shouldn't be a single boolean value to override the "opened" state for all parent techniques. My intention with this feature was that it'd set the `showSubtechniques` field for every TechniqueVM, therefore affecting exported layers and so forth as well. It'd essentially be a shortcut to the UX of clicking the sidebar on every technique.

nav-app/src/app/datatable/data-table.component.html
```diff
@@ -305,6 +305,16 @@
...
              </div>
 
+            <!-- show or hide disabled techniques -->
+            <div *ngIf="configService.getFeature('toggle_subtechniques')" class="control-row-item">
+                <div class="control-row-button noselect"
+                     (click)="viewModel.showAllSubtechniques = !viewModel.showAllSubtechniques;"
+                     matTooltipPosition="below"
+                     matTooltip="show/hide sub-techniques">
+                     <img [src]="viewModel.showAllSubtechniques ? 'assets/icons/label_24px.svg' : 'assets/icons/label_off_24px.svg'" [alt]="viewModel.showAllSubtechniques ? 'hide sub-techniques' : 'show sub-techniques'"/>
+                 </div>
```
isaisabel: This should probably be two different buttons. See my comment on the `showAllSubtechniques` field for more details, but we essentially need one button to set all `showSubtechniques` fields to `true` and one button to set them all to `false`.

A toggle like this _could_ work with that UX but it wouldn't be optimal -- if the user had one technique open and the rest closed, would clicking the toggle expand all or collapse all sub-techniques? The simpler UX is to simply have two buttons instead of a toggle.

Issue #112, opened 2019-11-12
Add subtechniques as configurable navigator feature


isaisabel:
As a user of the ATT&CK Navigator, I want to be able to remove subtechniques features from the view. This should use the existing "navigator features" functionality in the config file and the "create a customized navigator" feature.

See [this medium post](https://medium.com/mitre-attack/attack-sub-techniques-preview-b79ff0ba669a) for more information about subtechniques.

Issue #140, opened 2020-04-06
Make "select sub-techniques with parent" default to false


isaisabel:
The "select sub-techniques with parent" control's current default state (true) is not intuitive to users who overlook the fact that the behavior can be changed with the control. To allow users to more easily use the interface, the default state should be switched to "false" so that sub-techniques select independently by default.

PR #155, opened 2020-04-29
Added Sub-techniques as a Configurable Feature

clemiller:
Adds sub-techniques to the navigator features in the config file. When disabled, sub-techniques are not parsed and all relevant parts of the view are removed. Since this involved changes to the "select sub-techniques with parent" control, I also updated the default of the checkbox to false, which was requested in issue 140.

Resolves Issues #112 and #140

### DIFF
--- a/nav-app/src/app/config.service.ts
+++ b/nav-app/src/app/config.service.ts
@@ -18,7 +18,7 @@ export class ConfigService {
             //parse feature preferences from config json
             config["features"].forEach(function(featureObject: any) {
                 self.setFeature_object(featureObject);
-                if (featureObject.name === "subtechniques") {
+                if (featureObject.name === "sub-techniques") {
                     dataService.subtechniquesEnabled = featureObject.enabled;
                 }
             })
--- a/nav-app/src/app/config.service.ts
+++ b/nav-app/src/app/config.service.ts
@@ -18,22 +18,20 @@ export class ConfigService {
             //parse feature preferences from config json
             config["features"].forEach(function(featureObject: any) {
                 self.setFeature_object(featureObject);
-                if (featureObject.name === "sub-techniques") {
-                    dataService.subtechniquesEnabled = featureObject.enabled;
-                }
             })
             //override json preferences with preferences from URL fragment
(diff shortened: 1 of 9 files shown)


nav-app/src/app/config.service.ts
```diff
@@ -18,6 +18,9 @@ export class ConfigService {
             //parse feature preferences from config json
             config["features"].forEach(function(featureObject: any) {
                 self.setFeature_object(featureObject);
+                if (featureObject.name === "sub-techniques") {
+                    dataService.subtechniquesEnabled = featureObject.enabled;
+                }
```
isaisabel, line 23: This methodology causes a bug. 

There are two ways the user can change the feature-set of the Navigator: 
- Modify the structure in config.json manually.
- create a customized navigator link using the "create customized navigator" section on a new tab. When the Navigator loads it parses the query string (technically the fragment, but we use it like a query string) to set the initial state of the structure.

The options in the create customized navigator menu are automatically generated from the features structure and assuming the API is used properly there is no difference between the two options. By setting the controls in the code here, instead of using the API for checking if a feature is set, the user can't enable/disable navigator sub-techniques from the query string method.

Unfortunately it may not be possible to use the API as intended because that would make a dependency loop (config.service would be codependent on data.service, which IIRC is difficult to do).

This should be fixable by putting the dataService.subtechniquesEnabled setter to after the fragments have been parsed, and use `self.getFeature()` to get/set the value.

nav-app/src/app/data.service.ts
```diff
@@ -381,10 +384,12 @@ export class Technique extends BaseStix {
...
-        this.subtechniques = subtechniques;
-        for (let subtechnique of this.subtechniques) {
-            subtechnique.parent = this;
-        }
+        //if(dataService.subtechniquesEnabled) {
+            this.subtechniques = subtechniques;
+            for (let subtechnique of this.subtechniques) {
+                subtechnique.parent = this;
+            }
+        //}
```
isaisabel, line 392: The disabled code here should be removed and the indentation fixed.

Issue #162, opened 2020-05-12
Show number of sub-techniques under a given technique


isaisabel:
As a user of the ATT&CK Navigator, I want to be able to see the number of sub-techniques are under a given technique while the technique is collapsed. 

We should modify the technique cell format to include the sub-technique count suffix as used on the ATT&CK Website's matrices:

![image](https://user-images.githubusercontent.com/28868888/81717155-32428d80-9448-11ea-9786-a59f0200d630.png)

---

As a user, I also want to be able to tell how many of those sub-techniques have annotations while the parent is collapsed.

This can be done probably by modifying the suffix format. For example, 
- a technique with three sub-techniques, two of which are annotated, would be formatted as `(3/2)`. 
- A technique with 8 sub-techniques, all of which are annotated, would be formatted as `(8/8)`.
- A technique with 5 sub-techniques, none of which are annotated, would be formatted as `(0/5)`. This should also be deemphasized (displayed with medium/light gray text) so that the user's eyes are drawn to counts where sub-techniques have annotations.

Both counts should exclude disabled sub-techniques.

PR #176, opened 2020-06-17
Show number of sub-techniques/annotated sub-techniques

clemiller:
Adds a subscript on the technique cells to show the number of sub-techniques under a technique. The subscript also indicates how many of those sub-techniques have annotations and is deemphasized when none are annotated.

Resolves Issue #162


isaisabel:
This is excellent! The overall representation is great, but I have a few comments about some of the quirks of this implementation:
- Don't count disabled techniques when disabled techniques are hidden. This is done appropriately for the total count, but not the annotated count:

   Screen Shot 2020-06-19 at 9 33 40 AM
   when hidden: 
   Screen Shot 2020-06-19 at 9 33 50 AM
   See also #177.
- I really like the deemphasis of the sub-technique count when there are no annotations. That really helps the annotation count pop out to users when there are annotations.
- It appears that there is now a dark border-bottom on all technique cells that have sub-techniques. That wasn't there previously -- was this an intentional addition? Previously:
   Screen Shot 2020-06-19 at 9 37 11 AM
   Now:
   Screen Shot 2020-06-19 at 9 37 19 AM
   This actually matches the UI on the ATT&CK website better (where the entire super-technique cell has an outline, and the sub-techniques are better delimited):
   image
   However, if we make this change we should make sure the entire cell has a dark outline, instead of just the bottom.
- We should add a nonbreaking space before the subscript count to match [the representation on the ATT&CK Website](https://github.com/mitre-attack/attack-website/blob/develop/attack-theme/templates/macros/matrix.html#L233):
   image

isaisabel:
It appears that the border-bottom bug is related to the box-shadow applied to the element:

With the box shadow:
Screen Shot 2020-06-19 at 9 50 19 AM

Without the box shadow, the individual cells get their borders removed but the UI otherwise more closely resembles the attack-website representation.

Screen Shot 2020-06-19 at 9 50 21 AM

isaisabel:
I _think_ these bootstrap styles are what prevents the 1px container movement bug on the ATT&CK website:

image

clemiller:
That fixed the issue, thanks! I've removed the darker bottom border altogether.

Issue #199, opened 2020-07-09
Bug with multiple default layers in customized navigator instances 


isaisabel:
Using the "default layers" interface of the Create Customized Navigator feature, specifying two different default layers leads to the UI opening the 2nd specified layer twice. For example, specify the following two layers:

- `https://raw.githubusercontent.com/mitre-attack/attack-navigator/master/layers/data/update_layers/July_2020/July_2020_mobile_attack.json` 
- `https://raw.githubusercontent.com/mitre-attack/attack-navigator/master/layers/data/update_layers/July_2020/July_2020_enterprise_attack.json`

This generates the URL: `https://mitre-attack.github.io/attack-navigator/enterprise/#layerURL=https%3A%2F%2Fraw.githubusercontent.com%2Fmitre-attack%2Fattack-navigator%2Fmaster%2Flayers%2Fdata%2Fupdate_layers%2FJuly_2020%2FJuly_2020_mobile_attack.json&layerURL=https%3A%2F%2Fraw.githubusercontent.com%2Fmitre-attack%2Fattack-navigator%2Fmaster%2Flayers%2Fdata%2Fupdate_layers%2FJuly_2020%2FJuly_2020_enterprise_attack.json`

However, when the Navigator opens it actually loads the enterprise_attack layer twice.

isaisabel:
This has been fixed in the [feature/v4.0](https://github.com/mitre-attack/attack-navigator/tree/feature/v4.0) branch. It will be released on master with the upcoming v4.0 release.

Issue #200, opened 2020-07-10
Search on description returns no results unless you unclick Name and Attack ID under properties searched 


scampb06:
To reproduce the problem take the following steps.

1. Use Search Techniques within Selection Controls. Search for "ICMP" with all four "properties searched" selected. You get no results. 
2. Now unclick "Name" and "Attack ID" under "properties searched" and you will see two techniques - "Network Denial of Service:Direct Network Flood" and "Non-Application Layer Protocol".  
3. Similarly if you search for a specific "Attack ID" e.g. "T1095" with all four "properties searched" selected you get no results. 
4. When you unclick "Name" the technique "Non-Application Layer Protocol" appears.

isaisabel:
This has been fixed in the [feature/v4.0](https://github.com/mitre-attack/attack-navigator/tree/feature/v4.0) branch. It will be released on master with the upcoming [v4.0](https://github.com/mitre-attack/attack-navigator/milestone/5) release.

PR #210, opened 2020-07-29
Bug fix for multiple default layers

clemiller:
Provides a bug fix for the issue where specifying multiple default layers would open the last URL multiple times. Also fixes the bug in the "search techniques" interface where it would only return results from the first enabled search property.

Resolves issues #199 and #200


ION28:
Any ideas when this might get merged all the way up to master? I was also running into this issue in my project (we use the navigator to display "coverage" of ATT&CK across 3 different default layers) and was hoping to not have to make a fork.

isaisabel:
Hi @ION28,

This fix is part of our [v4.0 milestone](https://github.com/mitre-attack/attack-navigator/milestone/5), so you're welcome to track our progress there. I can't give a concrete release date but it probably won't be longer than a month or two.

Issue #207, opened 2020-07-20
Remove pre-ATT&CK domain


isaisabel:
With the next release of ATT&CK, the pre-ATT&CK domain is being migrated into new tactics under the enterprise domain. Therefore the pre-ATT&CK domain itself should be removed from the application. This also means that the "stages" (prepare/act) section of the filters and layer format may be removed.

To replicate the prior functionality of the stages filters, we may want to introduce the following feature to the context menu: 
- select/deselect all techniques in a tactic. This will allow users to easily select the new tactics in order to hide the pre-ATT&CK techniques using the existing disable/hide workflow.

PR #218, opened 2020-08-17
Removes pre-ATT&CK domain

clemiller:
Removes the pre-ATT&CK domain from the application, including:
- Removal of pre-ATT&CK urls from the config file
- Removal of "stages" in the "filters" interface and the layer format documentation
- Removal of data loading for pre-ATT&CK domain when updating outdated layers with the update-layers script
- Addition of option to select/deselect all techniques in a tactic, which can be done in both the context menu or by clicking the tactic name. These options will follow the user's selection preference under "selection behavior".

Resolves Issue #207


layers/update-layers.py
```diff
@@ -105,7 +99,7 @@ def update_layer(layerfile, replace=False):
             technique["techniqueID"] = newID
     
     # set the version to current
-    layer["version"] = "3.0"
+    layer["version"] = "4.0"
```
isaisabel, line 102: This script should also remove `layer["filters"]["stages"]` if it is present in the input json.

clemiller:
Hey Isabel, thank you for the suggestions for the tactic selection behavior issue! I've opted for solution 2: dynamically determining if all techniques in a tactic are selected or not, and cleaned up some of that code in the process.

clemiller:
Would there be a use case for not selecting the sub-techniques under the tactic? Changing the current logic would remove a users' ability to choose the "sub-technique selection" behavior when selecting a tactic, since it would always select the sub-techniques under the tactic whether the control has been selected or not. If that's not a concern, I can definitely make that logic change if it's more intuitive that way.

isaisabel:
Yeah that's why I said "I could go either way on [it]" -- I haven't thought of a use case where a user would want to select techniques in a tactic but _not_ sub-techniques, but there _could_ be one? There's certainly more user control over the behavior if we honor the "select sub-techniques with parent" control. 

One more argument in favor of keeping it the way it is (and not forcing sub-technique selection) is that the context menu prompt says "techniques" and not "techniques and sub-techniques." 

I guess my recommendation is that we can keep it the way it is, but make it clear in the documentation for the tactic selection behavior that it only selects sub-techniques if "select with parents" is enabled.

isaisabel:
The help page already says: 
>This action follows the behavior preference under selection behavior in the selection controls.

I think this needs to be more explicit about what "follows" means with respect to cross-tactic techniques and techniques with sub-techniques.

PR #223, opened 2020-09-03
Bugfix and enhance the render layer as SVG functionality

Che4ter:
In the new render layer as SVG version, the technique sort order was no longer honored.
This pull requests ads the sorting again.


Che4ter:
In the version 2.3.2 it was possible to hide the gradient in the svg export. The new version was missing this option, therefore I added it again.

isaisabel:
Hi @Che4ter,

Thank you for submitting this improvement. I had to make some additional changes in #232:
- Fixed a crash that could occur in certain layer configurations when the gradient was hidden in the exporter
- Legend display settings now only hide the user-defined legend, instead of the whole legend container. Previously disabling "show legend" hid the legend and scoring gradient since the whole box was hidden -- there was no way for the user to only hide the legend but not the scoring gradient. The legend container is now hidden if the user-defined legend and the scoring gradient are individually hidden.

PR #231, opened 2020-09-18
Feature/#212 adaptable platforms

### LINKED ISSUE #212: Adaptable platforms
As a user of the ATT&CK Navigator, I want to be able to use custom STIX data for my domain as part of the v4.0 milestone. Because flexible domains also include flexible data, the application can no longer make assumptions about what platforms may be found on the data. 

Currently platforms are hardcoded in the filter component depending on the domain the application is configured to use. In version 4.0, the platform options need to be detected from the input STIX data. Validation of uploaded layers should be performed against this dynamic list of platforms, and if the uploaded layer has a platform that is not detected on the relevant domain data, the user should be warned and the platform configuration ignored.

isaisabel:
Resolves #212: platform options are now loaded from the domain data instead of being hardcoded.

Note: I did encounter some errors when working with this, e.g when uploading layer files, but I'm pretty sure that's a bug in your branch since I didn't touch that code at all.

### DIFF
--- a/nav-app/src/app/data.service.ts
+++ b/nav-app/src/app/data.service.ts
@@ -4,6 +4,7 @@ import { HttpClient } from '@angular/common/http'
 import { Observable } from "rxjs/Rx"
 import { fromPromise } from 'rxjs/observable/fromPromise';
 import { TaxiiConnect, Server, Collections, Collection, Status } from './taxii2lib';
+import { platformBrowser } from '@angular/platform-browser';
 
 @Injectable({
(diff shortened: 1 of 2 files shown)


isaisabel:
Just verified that the upload layer bug is also present in your branch, so it doesn't look like that was from my changes here.

clemiller (approved):
Looks great!

Thank you for double checking that the upload layer bug exists in my branch, it'll be resolved before I create a pull request.

PR #236, opened 2020-10-14
Update the layer-updater script to use v4.0 of the Layer schema

### LINKED ISSUE #235: Revise update-layers to use Layer format v4.0
The upcoming layer format revisions in [version 4.0](https://github.com/mitre-attack/attack-navigator/milestone/5) need to get added to the update-layers script.

isaisabel:
Update the layer-update script to use v4.0 of the Layer schema.

Closes #235

### DIFF
--- a/layers/update-layers.py
+++ b/layers/update-layers.py
@@ -103,7 +103,11 @@ def update_layer(layerfile, replace=False):
             technique["techniqueID"] = newID
     
     # set the version to current
-    layer["version"] = "4.0"
+    layer["versions"] = {
+        "navigator": "4.0",
+        "layer": "4.0",
+        "attack": "8"
+    }
 
     # output layer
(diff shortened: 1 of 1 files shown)


clemiller:
The domain names are being refactored to `enterprise-attack` and `mobile-attack` as well,  which needs to be updated in the `domains` field at the top of the script. Additionally, should there be a way to validate that the outdated layer's `domain` field matches this standard?

isaisabel:
I'll update that. There will need to be backwards compatibility for the old domain schema (just enterprise and mobile) since all older layers will have the old schema. The backwards compatibility needs to be mentioned in the CHANGELOG if it isn't already.

I don't think layers are valid if their domain doesn't match the old/new standard, since knowing the domain is required for loading relevant data. It's a required field and there's no obvious default value for cases where the domain is malformed or missing.

clemiller:
Looks great! Thanks for updating the domain format!

Issue #245, opened 2020-10-30
Platforms filter is not working


marcusbakker:
Hi ATT&CK team,

When creating a layer file with a Filter object that has filters for specific ATT&CK platforms, it seems to be ignored when loaded into the Navigator.

If for example the attached file (`layer_network.json`) is loaded in the Navigator. All platforms are selected instead of only the `Network` platform.

[layer_network.json.zip](https://github.com/mitre-attack/attack-navigator/files/5467350/layer_network.json.zip)
```
{
	"name": "layer",
	"versions": {
		"attack": "8",
		"navigator": "4.0",
		"layer": "4.0"
	},
	"domain": "enterprise-attack",
	"description": "",
	"filters": {
		"platforms": [
			"Network"
		]
	},
	"sorting": 0,
	"layout": {
		"layout": "side",
		"showID": false,
		"showName": true
	},
	"hideDisabled": false,
	"techniques": [],
	"gradient": {
		"colors": [
			"#ff6666",
			"#ffe766",
			"#8ec843"
		],
		"minValue": 0,
		"maxValue": 100
	},
	"legendItems": [],
	"metadata": [],
	"showTacticRowBackground": false,
	"tacticRowBackground": "#dddddd",
	"selectTechniquesAcrossTactics": true,
	"selectSubtechniquesWithParent": false
}
```

<img width="158" alt="filters" src="https://user-images.githubusercontent.com/40516252/97751351-6ee76b80-1af2-11eb-87a0-65af33deaf6d.png">

Regards
Marcus

isaisabel:
This is an odd bug, it's definitely coded to deserialize the platforms[[1](https://github.com/mitre-attack/attack-navigator/blob/master/nav-app/src/app/viewmodels.service.ts#L1051), [2](https://github.com/mitre-attack/attack-navigator/blob/master/nav-app/src/app/viewmodels.service.ts#L1485-L1504)], but in testing it doesn't seem to actually apply them. 

This might be a bug from [loadVMData](https://github.com/mitre-attack/attack-navigator/blob/master/nav-app/src/app/viewmodels.service.ts#L405-L409), introduced when we changed the platform options to be dynamic (and not hardcoded). My guess is that `initPlatformOptions` is being called after the platforms are loaded from the layer, and are overwriting the selection. Further debugging is necessary to confirm if this is the cause.

isaisabel:
Fixed in the develop branch.

Issue #240, opened 2020-10-23
Indicate that a layer is loading when the user has uploaded a layer


isaisabel:
When importing a layer, the application takes a while to open the new tab because the data has to load first. The fact that loading is occuring should be indicated in the UI so the user knows the application is working. Otherwise users may conclude that an error has occurred or the layer indicated is not loading for some reason.

Issue #239, opened 2020-10-23
Indicate the version and domain when viewing a layer


isaisabel:
As a user of the ATT&CK Navigator, I want to be able to tell the version and domain I'm viewing when looking at a layer.
- Include in the layer UI somewhere (possibly layer information?)
- Include in exported SVGs as well?

Issue #189, opened 2020-06-30
Metadata layout in 3.0/3.1 version


rubinatorz:
Hi @isaisabel 

The layout of the metadata in the new 3.0/3.1 version of the Navigator is slightly changed compared to the 2.0 version. Values are now aligned with each other, the line height of each line is increased and the width of the tooltip is smaller:

<img width="366" alt="Screenshot 2020-06-29 at 17 13 00" src="https://user-images.githubusercontent.com/11735227/86110321-b9fe5d00-bac5-11ea-903a-cbf9637ca2cb.png">

In the previous version it was like this:
<img width="286" alt="Screenshot 2020-06-29 at 17 12 51" src="https://user-images.githubusercontent.com/11735227/86110364-c387c500-bac5-11ea-9ba7-5daa5416899f.png">

I can imagine the choice for this, but in our case it just looks less nice and a bit messy in the new version of the metadata. Is there a chance of having an option to influence this behaviour?

Thanks!

isaisabel:
Hi @rubinatorz,

Sorry, I did see your comment on the other issue yesterday I just haven't had time to respond until now.

The new tooltip implementation uses a borderless table instead of just a line by line listing. This theoretically improves readability although apparently not in your specific case. Two possible ways to improve the readability for your case:
1. remove the bullet points
2. Empty values could be represented as `""` instead of `"-"`. 
3. Remove the section breaks, or represent them as empty keys and values to make a blank line:
   ```json
    { "name": "", "value": ""}
   ```

One thing we could possibly change is setting white-space to `nowrap` for the lefthand side of the key-value pair. This way the metadata name wouldn't wrap when there are multiple words. In practice I'm not sure this actually improves things though:



![Screen Shot 2020-06-30 at 2 33 47 PM](https://user-images.githubusercontent.com/28868888/86164007-18840500-badf-11ea-89ca-747c3887b6f3.png)
above: add the following to line 13 of `nav-app/src/app/matrix/technique-cell/tooltip/tooltip.component.scss`:
```scss
        td:first-child {
            white-space: nowrap;
        }
```

![Screen Shot 2020-06-30 at 2 33 15 PM](https://user-images.githubusercontent.com/28868888/86164001-17eb6e80-badf-11ea-9b85-51b490ea8d7f.png)
above: how the same metadata looks with the current implementation.

[...]

PR #250, opened 2020-11-20
Mixed domains & versions v4.1

clemiller:
Features/improvements for version 4.1:
- Added a section break feature to the technique metadata format. Resolves issue #189. 
- Updated the layer file format to v4.1 to define the schema of the `divider` property.
- Added a layer's domain and version to the layer information interface and to the "about" section of the SVG exporter. Resolves issue #239. 
- Updated the UI to show a progress spinner when a user imports a layer to indicate the layer is loading. Resolves issue #240.
- Fixed the bug causing a layer's platform selection to be overwritten. Resolves issue #245.


layers/LAYERFORMATv4_1.md
```diff
@@ -0,0 +1,181 @@
...
+            "techniqueID": "T1053",
+            "tactic": "privilege-escalation",
+            "metadata": [
+                { 
+                    "name": "T1053 metadata1", 
+                    "value": "T1053 metadata1 value" 
+                },
+                {
+                    "divider": true
+                },
```
isaisabel, line 143: 👍  adding an example of divider usage

nav-app/src/app/datatable/data-table.component.html
```diff
@@ -116,6 +116,23 @@
...
+                                    [(ngModel)]="dataService.getDomain(viewModel.domainID).name" />
+                            </mat-form-field>
+
+                            <!-- layer version (uneditable) -->
+                            <mat-form-field>
+                                <input matInput readonly
+                                    type="text"
+                                    placeholder="version"
+                                    [(ngModel)]="dataService.getDomain(viewModel.domainID).version" />
+                            </mat-form-field>
```
isaisabel, line 134: These fields, though readonly, visually appear as if they're editable. Therefore it might be worth adding the `disabled` attribute, which should change them to be visually distinct while still following the overall pattern.

Example of what disabled looks like (note: it doesn't add the word `(disabled)` to the label, that's manually added in the example): 
image

nav-app/src/app/exporter/exporter.component.ts
```diff
@@ -478,6 +480,10 @@ export class ExporterComponent implements AfterContentInit {
             if (self.showName() || self.showDescription()) {
                 let about = {"title": "about", "contents": []};
                 if (self.showName()) about.contents.push({"label": "name", "data": this.viewModel.name});
+                if (self.showDomain()) {
+                    let domain = this.dataService.getDomain(this.viewModel.domainID);
+                    about.contents.push({"label": "domain", "data": domain.name + " " + domain.version})
+                }
```
isaisabel, line 486: The "about" panel is pretty full with name, domain/version, and description information. We might want to consider moving it over to the "filters" box which currently just has the platforms. We could rename filters to "showing" or something to that effect (definitely should try to come up with a better label than "showing" though).

image

nav-app/src/app/tabs/tabs.component.ts
```diff
@@ -545,22 +545,16 @@ export class TabsComponent implements AfterContentInit {
...
                 viewModel.deSerializeDomainID(string);
-
+                if (!this.dataService.getDomain(viewModel.domainID)) {
+                    throw {message: "Error: '" + viewModel.domain + "' (" + viewModel.version + ") is an invalid domain."};
+                }
                 this.versionUpgradeDialog(viewModel).then( () => {
-                    if (!this.dataService.getDomain(viewModel.domainID)) {
-                        throw {message: "Error: '" + viewModel.domain + "' (" + viewModel.version + ") is an invalid domain."};
-                    }
+                    this.openTab("new layer", this.layerTab, viewModel, true, true, true, true);
```
isaisabel, line 552: 👍  I like that it opens the tab first, then loads the data. Better than what I was originally thinking (putting the spinner over the new tab interface for the duration of the load).

isaisabel (changes requested):
A few minor notes below, also, the help page still links to the old layer PDF. 

We should consider adding #251 to this PR, or alternatively we can do it in a different branch/PR and merge it into develop for version 4.1. I haven't diagnosed #251 yet so I'm not sure how hard of a fix it's likely to be (browser compatibility fixes are usually either very easy or very finicky, there's not usually a lot of middle ground from my experience).

nav-app/src/app/exporter/exporter.component.ts
clemiller: I like the idea of moving it from the "about" section, so it'll be less cluttered. Since the "filters" section can also contain lot of content, what do you think of adding a new "domain" section to the header?
image

isaisabel: Fine by me! Probably should also add an option to hide it from the header though, some users might want to reclaim that horizontal space in small image sizes. 

@ArtificialErmine we'll need to make this change in the Python implementation.

isaisabel (approved):

PR #253, opened 2020-12-08
Upgrade to Angular 11, update other dependencies

### LINKED ISSUE #251: Loading layers via URL doesn't work on Safari
As a user of the ATT&CK Navigator, I want to be able to use my browser of choice without it impacting app performance or functionality.

Currently loading a layer by URL (potentially including default layers) on the Safari browser will hang the browser for an indefinite amount of time. Uploading a layer through the file browser appears to be unaffected.

isaisabel:
A (failed) attempt to resolve #251 by upgrading Angular and fixing dependencies. Still worth merging however since it supersedes #228, #226, #225, #211, #203, #173, #147, and #133.

Note: this still requires testing of the various interfaces. Initial tests did not discover any broken features.


clemiller (changes requested):
Most everything looks good here. The only feature that appears to break is the color picker in the "color setup" dropdown. The issue occurs on its initialization within the `ngx-color-picker` module itself: `cannot read property 'nativeElement' of undefined`

isaisabel:
Resolved by updating ngx-color-picker from version `7.5.0` to`10.1.0`.

I found only 1 bug in my testing, but I haven't been able to reproduce it -- I wasn't able to find one of my created layers in the `Create Layer from other layers` dropdown, even though the correct domain/version was selected. This was using Mobile v8. I was able to see a layer I created using a score expression _from_ that layer however, which is quite odd. 

Given that I can't reproduce it, I reckon we should just watch out for it in future testing and updates.

clemiller (approved):
I was also unable to reproduce that particular issue, but I'll keep an eye out for it. Everything looks good to me.

Issue #254, opened 2020-12-08
Refactor tabs implementation


isaisabel:
As a user of the ATT&CK Navigator, I want to be able to use the application without significant performance issues as a result of opening many tabs/layers.

The current implementation of the tabs interface has poor performance. Essentially the contents of each tab is its own DOM element, and while these elements are added/removed when tabs are switched for performance reasons, they are separate and the removal/addition of DOM elements can have poor performance. This is likely the cause of #251.

A new system should be developed to change how tabs work. Ideally:
- A data-table is always loaded. There are never multiple data-tables, that single data table will be used for displaying all layers.
- The "new tab" UI and "help" page are moved to modals/dialogs or some otherwise separate system, and are not represented in "tabs". The SVG export interface is likewise moved to a modal/dialog.
- Changing the "tab" changes the current ViewModel (layer) by changing which layer the data-table displays. 
    - the layer simply determines the current set of annotations and data-table settings (layout, gradient...). It also sets the domain/version data used to determine the table contents.
    - This may also change the current domain/version according to the domain/version of the layer. 
    - Some of this may happen asynchronously since data may have to be loaded when a layer is imported.

Something vaguely like this in terms of data flow:

![image](https://user-images.githubusercontent.com/28868888/101506754-be1b8a00-3943-11eb-8298-e45822090ce5.png)

isaisabel:
Addressed in #257.

Issue #179, opened 2020-06-23
Allow access to usage documentation outside of the application


isaisabel:
As a user of the ATT&CK Navigator, I want to be able to access the usage documentation (currently on the help page) outside of the application from my browser of choice. 

As a maintainer of the ATT&CK Navigator, I want to be able to link people asking to help to usage documentation on the ATT&CK Navigator repo, instead of saying "please see section `x` in the in-app help document."

---

Move the in-app help content into a markdown document, USAGE.md, in the root of this repo. Following the same method as #178, compile the usage doc into the in-app help document when the app builds.

isaisabel:
The documentation system for ATT&CK Workbench can be used as a basis for this feature.
- Markdown docs can be copied into the assets directory [using Angular](https://github.com/center-for-threat-informed-defense/attack-workbench-frontend/blob/develop/app/angular.json#L27).
- Markdown docs can be then be rendered in-app using [a markdown renderer](https://github.com/center-for-threat-informed-defense/attack-workbench-frontend/tree/develop/app/src/app/views/help-page).

isaisabel:
I'm marking this as a 3-pointer while #178 is only a 1-pointer because this one requires translating the current HTML docs into markdown, which may take some effort.

PR #257, opened 2020-12-30
Tabs refactor

clemiller:
Refactor of the tab/tabs component to resolve the performance issues caused by opening multiple tabs.
- Removal of the dynamic tabs directive
- A single data-table is used to display all of the layers; opening a "new" tab will hide the data-table
- Moved the help page and SVG exporter to a pop-up dialog window

Addresses issue #254


isaisabel (changes requested):
Your imports here are based on an outdated version of material from before the packages were updated. You should `npm install` to get the latest packages. 
The fix is that the import of dialog stuff in `help.component`.ts and `exporter.component.ts`:
```
import { MatDialogRef, MAT_DIALOG_DATA } from '@angular/material';
```
should be:
```
import { MatDialogRef, MAT_DIALOG_DATA } from '@angular/material/dialog';
```

Even despite this fix, I'm still having issues:
```
ng serve
? Port 4200 is already in use.
Would you like to use a different port? Yes
✔ Browser application bundle generation complete.

Initial Chunk Files | Names         |      Size
scripts.js          | scripts       |   2.70 MB
vendor.js           | vendor        | 340.34 kB
styles.css          | styles        |  74.76 kB
runtime.js          | runtime       |   6.15 kB
main.js             | main          |   1.80 kB
polyfills.js        | polyfills     | 683 bytes

                    | Initial Total |   3.11 MB

Build at: 2021-01-04T15:54:29.794Z - Hash: ebdc5e1d8a18ab77f0e5 - Time: 11649ms

Warning: Zone.js does not support native async/await in ES2017+.
These blocks are not intercepted by zone.js and will not triggering change detection.
See: https://github.com/angular/zone.js/pull/1140 for more information.



Error: src/app/tab/tab.component.ts:19:17 - error NG2003: No suitable injection token for parameter 'title' of class 'TabComponent'.
  Consider using the @Inject decorator to specify an injection token.

19     constructor(title: string, isCloseable: boolean, showScoreVariables: boolean, domain: string, dataTable: boolean) {
                   ~~~~~

  src/app/tab/tab.component.ts:19:24
    19     constructor(title: string, isCloseable: boolean, showScoreVariables: boolean, domain: string, dataTable: boolean) {
                              ~~~~~~
    This type is not supported as injection token.
src/app/app.module.ts:43:5 - error NG6001: The class 'TabComponent' is listed in the declarations of the NgModule 'AppModule', but is not a directive, a component, or a pipe. Either remove it from the NgModule's declarations, or add an appropriate Angular decorator.

43     TabComponent,
       ~~~~~~~~~~~~

  src/app/tab/tab.component.ts:8:14
    8 export class TabComponent {
                   ~~~~~~~~~~~~
    'TabComponent' is declared here.




** Angular Live Development Server is listening on localhost:55928, open your browser on http://localhost:55928/ **
```

Since I can't get past these errors, I can't actually review the UX/UI at all. One thing I am assuming from my skimming of the HTML changes is that there is no explicit close button for the help and exporter dialogs. While the user _can_ simply click off the dialogs to close 
them, they might not know this and think they're stuck in those UIs. Adding a close button at the top right or top left of the dialogs will alleviate this. Ideally the close button won't scroll with the dialog content so the user can see it wherever they are in the help page for instance.

I'll do a better review once the build issues are fixed.

isaisabel (changes requested):
Overall most of this looks pretty good. However, the SVG exporter really needs some polish. Currently it looks like this: 
image

1. There should be no padding around the toolbar at the top, _or_ the entire SVG area should have a grey-outlined box around it like the Navigator does. Currently it looks very strange that the toolbar doesn't extend to the full width of the dialog. Adding the border would also alleviate this issue I think.
2. The close button has some sort of bug when the dialog opens. I believe it is focused when the dialog opens, so it has a grey background until you click other elements of the dialog. It also has a _lot_ of whitespace around it because the dialog has built-in margins on the bottom. Removing the bottom margin from the button's div will make it visually nicer (and vertically centered in the whitespace).

Before (with extra margin):
Screen Shot 2021-01-11 at 9 57 15 AM
After (without extra margin):
Screen Shot 2021-01-11 at 9 57 22 AM

I'm sure you're aware of this, but #251 is still a problem (just switching quickly between two layer tabs eventually causes the freeze, I think that's the easiest way to reproduce it), and using the relative links in the help dialog closes the dialog (issue documented [here](https://stackoverflow.com/questions/65250123/how-to-use-html-anchor-link-id-inside-dialog-matdialog-in-angular-10) without a solution).

nav-app/src/app/help/help.component.scss
```diff
@@ -12,6 +12,12 @@
     .content {
         overflow-y: scroll;
         max-height: 60vh;
+
+        span {
+            cursor:pointer;
+            color:blue;
+            text-decoration:underline;
+        }
```
isaisabel, line 20: Probably this should also have a class so that spans used to apply other inline text effects won't look like hyperlinks. I don't think that happens right now but it would be future proofing for later additions to the help page. Also in #179 the markdown renderer may be more liberal in its application of <span>s (although this methodology may no longer be viable once #179 is implemented).

isaisabel (approved):
Looks great, though #263 needs to merge into this branch before we merge this one to develop.

PR #262, opened 2021-02-01
Remove safari support

clemiller:
Removes Safari support
- Adds a prompt to warn Safari users that the browser no longer supported
- Removes Safari from the help page and the list of supported browsers


isaisabel (approved):
Looks great!

Issue #265, opened 2021-02-13
Custom context variable substitution is not global


gclen:
First off, this project is amazing and thanks for all your hard work! When adding a custom context label only the first instance of a variable is substituted. For example if you had

```http://www.someurl.com/{{technique_attackID}}}/otherstuff/{{technique_attackID}}```

it would become

```http://www.someurl.com/T1234/otherstuff/{{technique_attackID}}```

not 

```http://www.someurl.com/T1234/otherstuff/T1234```

I have a pull request which fixes this coming soon.

isaisabel:
Addressed in #266 and merged into develop. Will be released to Master with Navigator v4.3.

PR #266, opened 2021-02-13
Add global substitution for context menu variables

gclen:
Fix for issue https://github.com/mitre-attack/attack-navigator/issues/265


isaisabel (approved):
Thank you for finding this bug and opening the PR! It looks great and will be included in the next release of the Navigator (version 4.3).

PR #276, opened 2021-03-24
Feature/note support

clemiller:
Adds support for note SDOs

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+# v4.3 - 26 February 2021
+
+- Added aggregate score that displays the combined score of the technique and all of its sub-techniques. See issue [#269](https://github.com/mitre-attack/attack-navigator/issues/269).
+- Added support for displaying note objects in tooltips.
+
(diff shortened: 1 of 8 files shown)


isaisabel (changes requested):
This works great as written, but could use a few additional quality of life features and documentation additions:

- CHANGELOG needs to be updated to mention note support
- README needs to be updated to mention STIX 2.1 support
   - Add mention of STIX 2.0 and 2.1 support to "loading content from local files"
   - Add mention that STIX 2.1/TAXII 2.1 is _not_ supported when loading from a TAXII server to  "loading content from a TAXII server'
- We should probably denote the presence of a note somehow in the UI outside of the tooltip.
   - Probably [add a "comment" underline](https://github.com/mitre-attack/attack-navigator/blob/master/nav-app/src/app/matrix/technique-cell/technique-cell.component.ts#L108-L136) style to technique cells which have notes, unless you can come up with a better alternative which can differentiate comments and notes. 
   - This should be documented in the help document somewhere.

isaisabel (approved):

Issue #270, opened 2021-03-05
Sub-techniques for T1584 and T1588 incorrect in generate Excel files


reginaelwell:
I've noticed that sub-techniques for T1584 and T1588 are incorrect in generate Excel files. Display in the Navigator and generated JSON files appear to be unaffected. 

T1584 has sub-techniques for T1583 in generated Excel files. 

T1584: Compromise Infrastructure
T1583.005: Botnet 
T1583.002: DNS   Server
T1583.001:   Domains
T1583.004:   Server
T1583.003:   Virtual Private Server
T1583.006: Web   Services

T1588 has a mix of sub-techniques for T1587 and T1588. 

T1588: Obtain Capabilities 
T1587.002: Code Signing   Certificates
T1587.003:   Digital Certificates
T1587.004:   Exploits
T1587.001:   Malware
T1588.002:   Tool
T1588.006:   Vulnerabilities

isaisabel:
Hi @reginaelwell,

This is an interesting bug. The objects in question have the same name, but different ATT&CK IDs. It appears that the excel exporter doesn't take the differing IDs into account and uses the wrong IDs (but correct name) when IDs are enabled under matrix layout. For example:

![Screen Shot 2021-03-08 at 10 47 35 AM](https://user-images.githubusercontent.com/28868888/110344645-b3809780-7ffb-11eb-8cbe-fbd77fc6c26f.png)

![Screen Shot 2021-03-08 at 10 50 37 AM](https://user-images.githubusercontent.com/28868888/110345019-1f630000-7ffc-11eb-8515-03073a64582f.png)

I'm 99% sure the bug occurs in [this line of code](https://github.com/mitre-attack/attack-navigator/blob/master/nav-app/src/app/datatable/data-table.component.ts#L134), where the sub-technique name is used as the key to lookup the ID. If this is the case, [line 148](https://github.com/mitre-attack/attack-navigator/blob/master/nav-app/src/app/datatable/data-table.component.ts#L148) may also have bugs associated with it (in the case of techniques with the same name, which shouldn't occur within the current dataset but could in the future).

As you noted, this only occurs in the excel exporter. The matrix layout and SVG exporter seem unaffected.

isaisabel:
This bug has been addressed in #279 and is now fixed on develop. It will be updated in the live application (on the master branch) with the next release.

PR #279, opened 2021-03-31
Bugs/#270 subtechniques with same names

iguannalin:
Fix for #270

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 - Added aggregate score that displays the combined score of the technique and all of its sub-techniques. See issue [#269](https://github.com/mitre-attack/attack-navigator/issues/269).
 - Added support for displaying notes in tooltips. Notes on a technique are indicated in the same style as a comment.
+- Fixed a bug in exporting matrix to Excel sheet, where the style of all sub-techniques with the same name in a column were incorrectly over-ridden by the style of the first sub-technique in its name. See issue [#270](https://github.com/mitre-attack/attack-navigator/issues/270).
 
 # v4.2 - 3 February 2021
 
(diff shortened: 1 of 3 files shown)


isaisabel (changes requested):
This seems to work quite well in resolving the original bug, but produces a new bug under a specific set of settings. 

Namely, techniques no longer honor the cell naming conventions after the first row. If under "matrix configuration" the label controls are adjusted, only the first technique cell in each column will have the proper label. For example: turn on only IDs for labels and only sub-techniques will display their ID, techniques (aside from the first of the column) will still show the name. Similarly for if name and ID are shown. It seems that sub-techniques are not effected by this bug, only techniques present the improper label.

Furthermore, no cells honor the label settings if both name and ID are turned off, although that bug is also present in the prior version of the Navigator (so it's not a new introduction of this branch).

iguannalin:
> This seems to work quite well in resolving the original bug, but produces a new bug under a specific set of settings.
> 
> Namely, techniques no longer honor the cell naming conventions after the first row. If under "matrix configuration" the label controls are adjusted, only the first technique cell in each column will have the proper label. For example: turn on only IDs for labels and only sub-techniques will display their ID, techniques (aside from the first of the column) will still show the name. Similarly for if name and ID are shown. It seems that sub-techniques are not effected by this bug, only techniques present the improper label.
> 
> Furthermore, no cells honor the label settings if both name and ID are turned off, although that bug is also present in the prior version of the Navigator (so it's not a new introduction of this branch).

Fixed with latest changes! Decided to revert to previous code, and make an array to keep track of already seen sub-techniques instead.
Not implementing the same in the techniques function, because all the technique cells within the column will have the same attackID, so it will thrown an error when styling the excel cells.
I thought about your idea of changing the techniqueCells array to hold a custom object for each technique's {name, id}, but it would mess with [Line 106](https://github.com/mitre-attack/attack-navigator/pull/279/files#diff-d74abd40373702f92538a0d1e1390240a72148fcaa8194c62e49a2162b66a2aaL106), where the array is edited in-place. The order of the items in the final array will matter when styling the excel sheet.

isaisabel (approved):
Works great with all configurations of cell display (name + ID)!

PR #280, opened 2021-04-03
Fix #278 - remove the use of fonts.googleapis.com

### LINKED ISSUE #278: Attack-navigator use fonts.googleapis.com
The attack-navigator app download the fonts from fonts.googleapis.com.

https://github.com/mitre-attack/attack-navigator/blob/5b6dbf918d186afaf899f1102305aceb8f4cc6c7/nav-app/src/app/tabs/tabs.component.html#L1

A good thing would be to download the fonts from the server of the attack-navigator app. This way, an "off-grid" attack-navigator app  won't need to get external sources to function properly.

Proposed fix (no tested):

```
npm install @fontsource/roboto
```
And update `/nav-app/src/styles.scss`
 ```
@import "~@fontsource/roboto/500.css";
```

2xyo:


isaisabel (changes requested):
Hi @2xyo!

Thanks for submitting this PR! We really appreciate your willingness to improve the application for the betterment of all our users.

I don't think this PR properly updates the fonts to use local versions. While it does no longer use network requests to load font resources, the UI has changed noticeably in a few places which indicates that fallback fonts are now being used. I'll include examples below and their likely causes:

## Case 1: new tab UI

I'm only 70% sure this is the cause, but most likely Angular Material doesn't like the fact that only roboto weight 500 is imported. Angular Material requires weights [300, 400 and 500](https://material.angular.io/guide/typography#usage) and you've only imported weight 500. So this _may_ be fixed if you import the other weights. 

At the very least Angular Material has stopped trying to load fonts from the web, indicating it's trying to read the fonts imported in this PR (but presumably failing since only one of the weights is present).

Live:
Screen Shot 2021-04-16 at 9 08 33 AM

This Branch:
Screen Shot 2021-04-16 at 9 08 40 AM

## Case 2: tab element UI

In this case this is because the monospace font used in most places in this app is [roboto-mono](https://fontsource.org/#roboto-mono). This PR installs roboto (not mono) which does not include the monospace variant. Installing and importing roboto mono will probably fix this.

Live:
Screen Shot 2021-04-16 at 9 10 04 AM

This Branch:
Screen Shot 2021-04-16 at 9 10 10 AM

2xyo:
Thank you for your great feedback and your patience!

Results of the following search https://github.com/mitre-attack/attack-navigator/search?q=roboto&type=code return the use of :

- `Roboto`  
  - https://github.com/mitre-attack/attack-navigator/blob/ae6e80fce75f6a95893a2b7f56ae64acebd6f6d8/nav-app/src/styles.scss#L7 
  - https://github.com/mitre-attack/attack-navigator/blob/ae6e80fce75f6a95893a2b7f56ae64acebd6f6d8/nav-app/src/app/matrix/matrix-side/matrix-side.component.scss#L80  
- `Roboto Mono` https://github.com/mitre-attack/attack-navigator/blob/ae6e80fce75f6a95893a2b7f56ae64acebd6f6d8/nav-app/src/app/tabs/tabs.component.scss#L21


So I added  `Roboto`  and `Roboto Mono` .

isaisabel (approved):
Hi @2xyo,

Sorry for the long delay in re-reviewing this. We've been very busy with the [ATT&CK Workbench](https://github.com/center-for-threat-informed-defense/attack-workbench-frontend) and [STIX 2.1 dataset](https://github.com/mitre-attack/attack-stix-data) release and I hadn't found time to revisit this until just now. 

Your changes look good and I'll be merging this into develop for inclusion in the next version of the application! Thank you very much for your contribution to this project.

Issue #252, opened 2020-11-23
Consolidate cloud platforms


isaisabel:
In the next release of ATT&CK, the `AWS`, `GCP`, and `Azure` platforms are consolidated into a single `IaaS` platform. The application should be updated to interpret any of these platforms as`IaaS` when loading layers from previous versions of ATT&CK. 

`Azure AD` is not affected by the merge.

isaisabel:
See also https://github.com/mitre-attack/attack-website/issues/276

PR #281, opened 2021-04-10
Features/#252 consolidate cloud platforms

iguannalin:
Feature add for [#252](https://github.com/mitre-attack/attack-navigator/issues/252).

Tested this feature by changing the config.json file to use latest version of STIX data as ATT&CK version 8 (to avoid errors), and imported an older data layer with the platforms "Azure", "GCP", and "AWS". The Navigator will then automatically deserialize the 3 platforms into the "IaaS" platform.


isaisabel (approved):
Looks great!

Issue #205, opened 2020-07-17
Add a configurable banner


isaisabel:
As a maintainer of the ATT&CK Navigator, I want to be able to add an optional announcement banner (like on [the ATT&CK website](attack.mitre.org)) to notify users when updates have been released. The content and visibility of the banner should be set in the config file.

PR #283, opened 2021-04-15
Added banner placeholder in tabs component, and added banner feature in config.json

iguannalin:
Feature for #205

### DIFF
--- a/nav-app/src/app/tabs/tabs.component.html
+++ b/nav-app/src/app/tabs/tabs.component.html
@@ -1,5 +1,12 @@
 <link href="https://fonts.googleapis.com/css?family=Roboto+Mono:500" rel="stylesheet">
 
+<!-- Banner -->
+<div *ngIf="hasFeature('banner') && bannerContent" class="header">
+    <div class="banner">
+        <span role="alert">{{ bannerContent }}</span>
+    </div>
+</div>
+
 <!-- Header/Help -->
 <div *ngIf="hasFeature('header')" [style.margin-bottom]="hasFeature('tabs') ? '0px' : '15px'" class="header" >
     <div class="logo">
--- a/nav-app/src/app/tabs/tabs.component.html
+++ b/nav-app/src/app/tabs/tabs.component.html
@@ -1,7 +1,7 @@
 <link href="https://fonts.googleapis.com/css?family=Roboto+Mono:500" rel="stylesheet">
 
 <!-- Banner -->
-<div *ngIf="hasFeature('banner') && bannerContent" class="header">
+<div *ngIf="hasFeature('header') && bannerContent" class="header">
     <div class="banner">
         <span role="alert">{{ bannerContent }}</span>
     </div>
--- a/nav-app/src/app/tabs/tabs.component.html
+++ b/nav-app/src/app/tabs/tabs.component.html
@@ -3,12 +3,20 @@
 <!-- Banner -->
 <div *ngIf="hasFeature('header') && bannerContent" class="header">
     <div class="banner">
-        <span role="alert" [innerHTML]="bannerContent"></span>
+        <span class="bannerContent" role="alert" [innerHTML]="bannerContent"></span>
+        <div [style.margin-bottom]="hasFeature('tabs') ? '0px' : '15px'" class="bannerLinks">
+            <div>
+                <a href="http://attack.mitre.org">MITRE ATT&CK&reg; Navigator</a>
+            </div>
+            <div *ngIf="hasFeature('tabs')" class="helpIcon" matTooltip="help" (click)="openHelpDialog()">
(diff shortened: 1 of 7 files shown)


iguannalin:
@isaisabel I spent some time exploring the banner UI, but before I continue I had a few questions I wanted to clarify about where the content for the banner is coming from, and what it should contain.
Right now, I have it set to show the content inside a field for the new config feature setting 'banner' [here](https://github.com/mitre-attack/attack-navigator/blob/7a311409751a5722e1a0c82dd048bf2d0afa2e42/nav-app/src/assets/config.json#L89)--does this look right?

isaisabel:
No, it shouldn't be stored in the features structure. The reason being that the features structure is also used to populate the UI in the _create a customized navigator_ section of the new tab, and I don't think the banner settings should be in there.

That said, the _header_ setting of the feature struct _should_ remove the banner and basically anything else outside of the Navigator "box" and tabs. The _header_ setting is useful for when users embed the Navigator in an iframe, which looks weird when there's those elements.

As for where the feature should go, a new field at the top level of the config file, like `"banner": "string"` (adjacent to `"comment_color"`), would probably be the best way to do this. If the banner string was empty it would not be rendered.

nav-app/src/app/tabs/tabs.component.html
```diff
@@ -1,5 +1,12 @@
 <link href="https://fonts.googleapis.com/css?family=Roboto+Mono:500" rel="stylesheet">
 
+<!-- Banner -->
+<div *ngIf="hasFeature('header') && bannerContent" class="header">
+    <div class="banner">
+        <span role="alert">{{ bannerContent }}</span>
```
isaisabel, line 6: At present this will sanitize any HTML in the banner, which means we can't put a URL or formatting in the banner. URLs are useful since they would allow use to link to a changelog or release notes for instance.

I believe Angular allows un-santitized HTML via some means (I believe the color gradient code may do this already?), but another option would be to use [ngx-markdown](https://github.com/jfcere/ngx-markdown) to render the banner content allowing us to use markdown formatting in the banner content.

isaisabel (changes requested):
Functionally I think this works fine in terms of how to configure it, with the caveat that it would be nice to be able to support formatting and hyperlinks in the banner content (see comment below).

Stylistically, two notes:
- If the banner is not going to reach the full width of the viewport, it should at least line up with the table container. Currently it's a few pixels off:
   Screen Shot 2021-04-15 at 10 26 22 AM
- Ideally the banner and the rest of the header content (app title and help page link) would be in a single row. Currently the banner ends up being rendered before the title and to the left of the help button, which is firstly not terribly aesthetic and secondly not a good use of space. Perhaps something like this (when the banner is present) would be better? 
   banner_example
   This may require some changes to the header layout when a banner is _not_ present, but the header has always been a bit wonky so that might be worth doing anyway.

isaisabel:
Also we should probably mention the capability in the root README somewhere. Probably near the "Disabling Navigator Features" section which is where all the customizing/config docs live.

nav-app/src/app/tabs/tabs.component.scss
```diff
@@ -188,6 +188,43 @@ input[type=file]{
...
+        .helpIcon {
+            height: 18px;
+            padding-left: 15px;
+        }
+        ::ng-deep a, a, ::ng-deep * {
+            display: inline;
+            color: on-color(primary);
+            text-decoration-line: none;
+            text-decoration-color: on-color(primary);
+        }
```
isaisabel, line 222: This removes any text styling for hyperlinks, which breaks the banner if we wanted to embed multiple links. 
For instance:

`&#9888; ATT&CK version 9 has been released! See <a href=\"google.com\">the changelog</a> or <a href=\"google.com\">visit the website</a>.`

> &#9888; ATT&CK version 9 has been released! See <a href="google.com">the changelog</a> or <a href="google.com">visit the website</a>.



Which displays as:

image

The only way to tell where the hyperlinks are is by hovering with the mouse. Ideally `a` elements should be styled somehow -- maybe keeping the underline? It's a bit funky because of the orange background.

isaisabel: I believe `::ng-deep` is a deprecated feature. The correct way to do this in recent versions of Angular is through [encapsulation](https://github.com/mitre-attack/attack-navigator/blob/master/nav-app/src/app/datatable/data-table.component.ts#L26).

isaisabel (changes requested):
Minor stylistic issues but feature wise works great!

In addition to my comment about hyperlink visibility, there is an issue where the entire header has a cursor:pointer style applied to it, indicating clickability where none exists. This probably comes from line 162 of `tabs.component.scss`.

isaisabel (changes requested):
The most recent update doesn't seem to address the issue with links not being apparent in the banner (https://github.com/mitre-attack/attack-navigator/pull/283#discussion_r621436492), although all the other changes requested were made.

isaisabel (approved):

Issue #256, opened 2020-12-22
Expand sub-techniques where annotations exist


isaisabel:
As a user, I want to be able to expand only the techniques which have annotated sub-techniques, instead of expanding _all_ techniques. This would allow me to find hidden annotations easily while keeping un-annotated sub-techniques out of the view.

PR #285, opened 2021-04-16
Features/#256 expand annotated subtechniques

iguannalin:
Feature for #256


isaisabel (changes requested):
I think this is a misinterpretation of the #256. You can already achieve very close to the functionality of this PR by:
1. select unannotated (in context menu)
2. disable the resultant techniques
3. turn on hide disabled techniques

This feature was intended to cover the expansion of the parent technique only when sub-techniques are present with annotations. Currently there is a button in the menu to expand all techniques with sub-techniques. #256 asks to add another button to only expand techniques whose sub-techniques have annotations. Techniques with sub-techniques which _don't_ have annotations would be left un-disabled/un-hidden, but wouldn't be expanded.

So this wouldn't require changes to the layer file format or application state, it's just a new button to expand specific sets of techniques easier.

PR #287, opened 2021-04-22
Features/#256 expand annotated subtechniques

iguannalin:
Re-opened PR from #285


nav-app/src/app/datatable/data-table.component.ts
```diff
@@ -285,13 +285,26 @@ export class DataTableComponent implements AfterViewInit {
...
-                    tvm.showSubtechniques = true;
+                    if (!showAnnotatedOnly) {
+                        tvm.showSubtechniques = true;
+                    } else {
+                        let foundAnnotated = false;
+                        for (let subtechnique of technique.subtechniques) {
+                            for (let sid of subtechnique.get_all_technique_tactic_ids()) {
+                                let svm = this.viewModel.getTechniqueVM_id(sid);
+                                if (svm.annotated() && !foundAnnotated) {
+                                    foundAnnotated = true; // set boolean after it finds the first annotated subtechnique
```
isaisabel, line 302: I believe you can `break` here for some added efficiency. Either that or collapse the whole thing into `subtechnique.get_all_technique_tactic_ids().some((sid) => { ... })` [which probably does the same thing](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some).

nav-app/src/app/datatable/data-table.component.ts
```diff
@@ -285,13 +285,26 @@ export class DataTableComponent implements AfterViewInit {
...
+                                }
+                                tvm.showSubtechniques = foundAnnotated;  // otherwise, keep going and collapse if no annotations found
```
isaisabel, line 304: This actually isn't the expected behavior. My expectation, according to the iconography and behavior of the counterpart "expand all" control, is that the button wouldn't affect techniques that were already expanded. 

The fact that it collapses all unannotated techniques isn't quite what I was expecting here, and limits the user (they may want to keep their current set of expanded techniques and also expand ones with annotations). The functionality of only expanding ones with annotations can be replicated by collapsing all and then clicking the button, so we don't really gain any functionality from this behavior.

The fix would be to make this line into `tvm.showSubtechniques == tvm.showSubtechniques || foundAnnotated` so that it keeps its expanded state it it was already set.

isaisabel (changes requested):
Found one minor behavioral quirk and one easy performance improvement. Should be quick fixes.

Also, the feature should be documented in the help page. There's an existing section for those controls so it should just be a new bullet point.

isaisabel (approved):

PR #289, opened 2021-05-05
Feature/#181 layer upgrade UI update

iguannalin:
Moved sidebar to inside matrix layout, rather than outside as overlay on top of app

### DIFF
--- a/nav-app/src/app/app.component.html
+++ b/nav-app/src/app/app.component.html
@@ -3,7 +3,4 @@
         <tabs></tabs>
         <span style="font-size: 7pt">MITRE ATT&CK&reg; Navigator v{{nav_version}}</span>
     </mat-drawer-content>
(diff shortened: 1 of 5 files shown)


clemiller (changes requested):
This looks great overall! It's definitely an improvement from having it overlay the entire application. Just one note: we may want to move the `mat-drawer-container` and `mat-drawer-content` tags from the app component to the data table component as well, and display the matrix views inside of the `mat-drawer-content`. Currently, the sidebar overlaps the right side of the data table, which blocks a number of techniques from the user's view, and doesn't extend down to the bottom of the page. This might require some additional modifications, such as setting the overflow of the `mat-drawer-content` when the sidebar is open so the user can scroll through the datatable, and preventing the sidebar from closing when the user clicks outside.

clemiller (approved):
This looks great!

Issue #204, opened 2020-07-17
Consolidate Search and Multiselect Interfaces


isaisabel:
As a user of the ATT&CK Navigator, I want to be able to explore the knowledge base in an intuitive interface. Currently, it's not possible to search inside of the multiselect interface for particular groups, software, or mitigations. Consolidating the search and multiselect UI into a single interface would improve the functionality of both UIs and the quality of the overall application.

![image](https://user-images.githubusercontent.com/28868888/87804628-22697000-c822-11ea-99b3-d2d3e19ab860.png)

![image](https://user-images.githubusercontent.com/28868888/87804636-24cbca00-c822-11ea-807c-e979b73edce1.png)

![image](https://user-images.githubusercontent.com/28868888/87804645-27c6ba80-c822-11ea-8a22-1d14515c1e48.png)

![image](https://user-images.githubusercontent.com/28868888/87804648-2ac1ab00-c822-11ea-8d02-85f981031012.png)

![image](https://user-images.githubusercontent.com/28868888/87804654-2d240500-c822-11ea-8faa-8e993342342d.png)

![image](https://user-images.githubusercontent.com/28868888/87804660-2eedc880-c822-11ea-9aaa-3213b72db151.png)

![image](https://user-images.githubusercontent.com/28868888/87804676-33b27c80-c822-11ea-8276-e51ef6035db0.png)

isaisabel:
We may also want to consider whether #150 can be addressed in this feature as well.

isaisabel:
Should also have a section for Notes (see #276), so that users can see listed notes, and select techniques with the notes. The display for notes may need to be different than the other types since notes have titles and descriptions.

isaisabel:
Move sidebar to the right side instead of left (as shown in mockups) since lefthand sidebars are more commonly used for navigation elements.

PR #301, opened 2021-06-09
Feature/#204 consolidate search and multiselect

iguannalin:
See issue: [#204](https://github.com/mitre-attack/attack-navigator/issues/204)


nav-app/src/app/sidebar/sidebar.component.ts
```diff
@@ -0,0 +1,18 @@
+import { Component, Input, OnInit } from '@angular/core';
+import { DataService } from '../data.service';
+import { ViewModel } from '../viewmodels.service';
+
+@Component({
+    selector: 'sidebar',
+    templateUrl: './sidebar.component.html',
+    styleUrls: ['./sidebar.component.scss']
```
isaisabel, line 8: This technically violates the custom component prefix recommendation for Angular, but so do _most if not all_ of the Navigator custom components so it's probably not worth fixing in just this one case. We may want to do a short code cleanup sprint later to fix this.

https://angular.io/guide/styleguide#component-custom-prefix

nav-app/src/app/techniques-search/techniques-search.component.ts
```diff
@@ -1,6 +1,6 @@
 import { Component, OnInit, Input } from '@angular/core';
 import { ViewModel } from '../viewmodels.service';
-import { DataService, Technique } from '../data.service';
+import {BaseStix, DataService, Group, Mitigation, Software, Technique} from '../data.service';
 
 @Component({
   selector: 'app-techniques-search',
```
isaisabel, line 6: We may want to rename this component from `techniques-search` to something more general since we can search beyond just techniques in the interface now.

nav-app/src/app/techniques-search/techniques-search.component.scss
```diff
@@ -48,4 +67,7 @@
             text-align: center;
         }
     }
+    ::ng-deep mat-panel-description {
+        align-items: center;
+    }
```
isaisabel, line 72: I believe the correct way to approach this is to disable ViewEncapsulation, ::ng-deep is deprecated:https://angular.io/guide/component-styles#deprecated-deep--and-ng-deep

nav-app/src/app/techniques-search/techniques-search.component.scss
```diff
@@ -1,11 +1,25 @@
...
-        padding: 0 10px;
+    ::ng-deep .mat-form-field-appearance-outline {
+        width: 100%;
+    }
+    ::ng-deep .mat-form-field-outline-start {
+        width: 30px !important;
+    }
+    ::ng-deep .mat-form-field-outline-start, ::ng-deep .mat-form-field-outline-gap, ::ng-deep .mat-form-field-outline-end {
+        background-color: white;
+    }
```
isaisabel, line 11: ::ng-deep is deprecated (see other comment about this)

nav-app/src/app/techniques-search/techniques-search.component.ts
```diff
@@ -37,71 +39,140 @@ export class TechniquesSearchComponent implements OnInit {
...
+    ngOnInit() {
+        this.getStixData();
+        this.getTechniques();
+    }
+
+    // filterAndSort() takes an array of items and does the following:
+    // 1) if the query is empty, then it sorts the array
+    // 2) if the query is not empty, then it filters the already sorted array until nothing is left, or,
+    // the query is cleared out and empty again
+    filterAndSort(items: any[], query = "", sortAllTechniques = false) {
```
isaisabel, line 70: Args should probably be documented in doc comment, it's not inherently apparent what sortAllTechniques does without a doc comment.

nav-app/src/app/datatable/data-table.component.html
```diff
@@ -41,32 +41,13 @@
             <div *ngIf="configService.getFeature('search')" class="control-row-item">
 
                 <div class="control-row-button dropdown noselect"
```
isaisabel, line 43: The dropdown class should be removed here (which _should_ remove the arrow after the button) since it's not really a dropdown anymore.

nav-app/src/app/techniques-search/techniques-search.component.html
```diff
@@ -1,34 +1,102 @@
...
+            <div>
+                <div class="allresults-buttons">
+                    <button class="button" (click)="selectAll(techniqueResults)">select all</button>
+                    <button class="button" (click)="deselectAll(techniqueResults)">deselect all</button>
+                </div>
+                <div class="results">
+                    <table class="results" *ngIf="techniqueResults.length > 0">
+                        <tr *ngFor="let result of techniqueResults">
+                            <td (mouseenter)="mouseEnter(result)" (mouseleave)="mouseLeave()">
+                                <span *ngIf="result.isSubtechnique">{{result.parent.name}}: </span>{{result.name}}
```
isaisabel, line 49: Glad you kept the parent name as a prefix to the sub-technique name. I don't think I've mentioned it previously, but that's a standard we keep across the infrastructure -- if not shown hierarchically (not in the matrix or indented list, etc), the sub-technique name is always prepended with the parent name.

nav-app/src/app/techniques-search/techniques-search.component.ts
```diff
@@ -37,71 +39,140 @@ export class TechniquesSearchComponent implements OnInit {
...
+        let self = this;
```
isaisabel, line 71: It seems like deprecated items are not being filtered out of the sidebar results:
image
"bash_profile and .bashrc Mitigation" is a deprecated mitigation and shouldn't be shown.

nav-app/src/app/techniques-search/techniques-search.component.html
```diff
@@ -1,34 +1,102 @@
 <div class="techniques-search">
-    <div class="query">
-        <mat-form-field style="width:100%; margin-top: 10px; box-sizing: content-box;">
-            <input matInput [(ngModel)]="query" placeholder="Search Techniques">
-        </mat-form-field>
+    <mat-form-field appearance="outline">
+        <mat-label>Search Techniques</mat-label>
```
isaisabel, line 3: Label here should not be "search techniques" since it's searching all object types, not just techniques. Probably just "search" would be appropriate?

nav-app/src/app/techniques-search/techniques-search.component.ts
isaisabel: A quick way to tell (for mitigations specifically) is that if it ends with the word "mitigation" and the prefix looks like a technique name it's almost definitely a deprecated mitigation. There was an update which changed how mitigations behave and they used to follow that naming convention (`"[technique] Mitigation"`)

nav-app/src/app/techniques-search/techniques-search.component.ts
isaisabel: Should `items` be `BaseStix[]` instead of `any[]`? It seems like it will always be a type which inherits from [BaseStix](https://github.com/mitre-attack/attack-navigator/blob/master/nav-app/src/app/data.service.ts#L308).

nav-app/src/app/techniques-search/techniques-search.component.ts
```diff
@@ -37,71 +39,140 @@ export class TechniquesSearchComponent implements OnInit {
...
-    ngOnInit() {
+    // getStixResults() checks if query is:
+    // 1) valid, and
+    // 2) part of last query, otherwise call getStixData() to search all objects again
+    getStixResults(query = "") {
+        if (query.trim() != "" && query.includes(this.previousQuery)) {
+            this.stixTypes.forEach(item => item['objects'] = this.filterAndSort(item['objects'], query));
+        } else {
+            this.getStixData();
+        }
```
isaisabel, line 144: I've not seen optimization like this before. Definitely something I will consider reusing myself in the future!

nav-app/src/app/techniques-search/techniques-search.component.ts
```diff
@@ -37,71 +39,140 @@ export class TechniquesSearchComponent implements OnInit {
...
 
+    public mouseEnter(technique: Technique, isStix?: boolean): void {
+        if (isStix) {
+            for (let t of this.getRelated(technique)) {
+                this.viewModel.selectTechniqueAcrossTactics(t, true, true);
+            }
+        } else {
+            this.viewModel.highlightTechnique(technique);
+        }
+    }
```
isaisabel, line 171: For the purposes of correctness `isStix` should be changed. Techniques are themselves STIX like all other objects. `isNonTechnique` (or `isTechnique` while inverting the if statement) would be better.

nav-app/src/app/techniques-search/techniques-search.component.ts
```diff
@@ -110,12 +181,50 @@ export class TechniquesSearchComponent implements OnInit {
...
+        for (let stixObject of items) {
+            this.selectStix(stixObject);
+        }
+    }
+
+    public deselectAllSitx(items: BaseStix[]): void {
+        for (let stixObject of items) {
+            this.deselectStix(stixObject);
+        }
+    }
```
isaisabel, line 214: Again I think this is a misuse of the word Stix. Also STIX is misspelled in `deselectAllSitx`

isaisabel (changes requested):
This is an _extremely_ snazzy UI. I am thoroughly impressed! The elevation of the blocks on the gray background is an excellent way to section the items and draw the eye to the important parts while keeping the sidebar visually separated from the matrix.

# Behavior
Behaviorally I have not found any notable bugs with what is selected. All of the edge cases I tried (involving the selection controls) checked out just fine. I have no concerns about the overall behavior of the feature.

We probably want to include an explicit close button on the sidebar since users might not know they can click the search button to close the sidebar. Maybe making it to the right of the search input (top right corner), inline with the input?

# Performance
I am a little bit concerned with the performance of the new highlighting code. Some quick performance analysis on the previous code shows it takes about 49ms to execute the mouseenter event and callbacks to highlight a cell. The new code takes anywhere from 70 ms to 100ms, which is _visibly_ less responsive. 

## Previous mouseenter performance:
Screen Shot 2021-06-09 at 3 45 02 PM

## Current mouseenter performance:
Screen Shot 2021-06-09 at 3 45 53 PM

We may want to look into possible performance enhancements to avoid visual latency in that part of the application. 

I should note that performance is always slightly better in the actual live deployment (presumably because of optimizations in the production build), so we may want to do more testing of the performance just in case I'm wrong in my method of measurement. Probably the best way to test performance is to check out the previous version (4.3) and run it in parallel on a different port and take the average of 10 observances of mousenter or something to that effect. This will produce closer conditions than in my testing.

nav-app/src/app/sidebar/sidebar.component.ts
iguannalin: Yeah, that's a good point. @clemiller and I talked about this today, and I think we should wait until both our issues have been merged into the develop branch before renaming the sidebar component.

iguannalin:
> We may want to look into possible performance enhancements to avoid visual latency in that part of the application.
> 
> I should note that performance is always slightly better in the actual live deployment (presumably because of optimizations in the production build), so we may want to do more testing of the performance just in case I'm wrong in my method of measurement. Probably the best way to test performance is to check out the previous version (4.3) and run it in parallel on a different port and take the average of 10 observances of mousenter or something to that effect. This will produce closer conditions than in my testing.

@isaisabel, I looked into this and tested the performance of the _mouseenter_ event in 3 environments:
1. locally, develop branch (65.78ms)
2. locally, master branch (78.78ms)
2. locally, feature/#204 branch (76.59ms)
3. deployed[ v4.3](https://mitre-attack.github.io/attack-navigator/) (20.08ms)

I tested this by running the performance profiling, reloading the page, and hovering over the 4 techniques in the top left corner after the matrix loads. It seems to take consistently ~60-80ms for the mouseenter event when running locally (I ran these about 2-3 times), and significantly less at 20ms in the deployed version on the mitre-attack site. 

If I'm doing this right and understanding you correctly, I'm not sure if there's a significant lag in the mouseenter/hover because of the changes in this feature branch. One thing I do think concerns me about performance when working on this feature though, was the loading of all the techniques in the search component sidebar (since it now loads & searches techniques across domains and all techniques).

Screenshots:
1. Screen Shot 2021-06-16 at 9 32 16 AM
2. Screen Shot 2021-06-16 at 9 32 33 AM
3. Screen Shot 2021-06-16 at 9 39 24 AM
4. Screen Shot 2021-06-16 at 9 31 10 AM

isaisabel (changes requested):
I actually have found one major bug in my second review. When switching between layers of different domains, the contents of the sidebar does not update properly until a query is added.

**Steps to reproduce:**
1. Open two layers, one Enterprise and one Mobile
2. Open the sidebar on both tabs
3. Switch back and forth between the two tabs. The counts and listed objects should change when switching tabs since the data under each domain is different, but this does not happen. Only when a query is inputted do the objects listed refresh.

Aside from that everything looks great, and the added denounce definitely makes the search input more performant.

isaisabel (approved):
Looks great! I pushed a commit which updates the help page documentation which was pretty old for this feature (it didn't mention mitigations which were not originally part of the interface and generally needed a touch-up).

Issue #295, opened 2021-05-12
Sort by highest score appears to be broken


beerMT:
I love the roll-up (sum) of SubTechniques into the Techniques score, thanks for that update!

When sorting by most frequently occurring Technique, the following was the result:
<img width="1864" alt="sorted_butNot" src="https://user-images.githubusercontent.com/6117308/117991332-2e677680-b2fb-11eb-8b59-af7bfab4d11c.png">

I have "hotter" red items that did not bubble up. Any help here on regaining that functionality would be great.

Cheers,
Brad

isaisabel:
Hi @beerMT,

Is there any chance you can upload the layer you're using so that we can use it to debug? I'm having a bit of trouble reproducing the issue.

beerMT:
Thanks for taking a look. I have attached a test merged layer, which consists of roughly 33 TA's merged together that target a test industry. The Techniques/Subtechniques were added together/incremented if present in the jsons. This was the layer that produced the output heatmap for opening the issue.

[merged_test.json.zip](https://github.com/mitre-attack/attack-navigator/files/6495468/merged_test.json.zip)

iguannalin:
Hi @beerMT,

Thanks for sharing that json file. You're right, I see that the colors are being calculated correctly, but it does not seem to be sorting by the aggregate score correctly when that's enabled. We should create this as an issue.

isaisabel:
This issue is now fixed on develop and will be released with Navigator v4.4.

PR #305, opened 2021-06-10
Fix/#295 sort techniques by aggregate score

iguannalin:
Fixed issue [#295](https://github.com/mitre-attack/attack-navigator/issues/295), example JSON file to test can be found in thread.


isaisabel:
This supersedes #304, right? They look mostly identical.

iguannalin:
@isaisabel Yep, that's a good point, I had time to work on this one, so I thought I'd push this. The two should not have too much overlap though, so I don't there will be merge conflicts.

nav-app/src/app/viewmodels.service.ts
```diff
@@ -1130,6 +1130,7 @@ export class ViewModel {
             else console.error("TypeError: sorting field is not a number")
         }
         if ("hideDisabled" in obj) {
+            obj.hideDisabled = JSON.parse(obj.hideDisabled);
```
isaisabel, line 1133: I'm curious why this (and the below uses of `JSON.parse` for booleans) is nessecary? `deSerialize` already parses the entire structure, so this would only be useful if the boolean here was a string (`"false"`) instead of an actual boolean (`false`). "String booleans" aren't technically allowed in the Layer spec so I don't see why it should be added here unless there's something I'm missing.

isaisabel (commented):
Definitely resolves the bug, but I'm curious about one of the additions that _seems_ unrelated.

isaisabel (approved):

isaisabel (changes requested):
Sorry, realize I already approved it but I just noticed there is no changelog entry for this. Aside from that it's all set!

isaisabel (approved):

Issue #298, opened 2021-05-20
Active sub-techniques are not displayed after filtering 


peterkimbler:
## Reproducible Steps

- https://mitre-attack.github.io/attack-navigator/
- Create New layer ->  Enterprise
- Click "Expand sub-techniques"
- Select "Active Scanning (T1595)" and "Botnet (T1583.005)"
- Right click -> Invert Selection
- Click "Toggle state"
- Click "Show/Hide disabled"

## Actual Output

![image](https://user-images.githubusercontent.com/72859632/118948048-37d68b80-b958-11eb-8cdf-16a89231dadb.png)

## Expected Output

"Active Scanning (T1595)" and "Botnet (T1583.005)" displayed in two columns.

anig1scur:
A workaround is to select "Acquire Infrastructure(T1583)"

It seems that there is no way to show a sub-technique in single column, which needs parent-technique to hold on.

isaisabel:
Yes I think this is sort of expected behavior -- if the parent technique is hidden, so are all sub-techniques. We may want to consider making a change to this behavior however since it can lead to data the user didn't intend to hide.

Written as a user story:
> As a user, I want to be able to see enabled sub-techniques even if their parents are disabled and hide disabled techniques is enabled. In this case the parent technique should be also be shown for context, but with the usual disabled style applied to denote that it is disabled and would otherwise be hidden.

isaisabel:
This has been added in #310 and is now present on the develop branch. It will be deployed to the live instance in version 4.4.

PR #310, opened 2021-06-23
Feature/#298 show active subtechniques

iguannalin:
Feature add for #298; to show active subtechniques even if parent technique is inactive


isaisabel:
@iguannalin this should target the `develop` branch, not `master`. It looks like it can't currently be retargetted to that branch since it has no commits that aren't on develop yet.

iguannalin:
> @iguannalin this should target the `develop` branch, not `master`. It looks like it can't currently be retargetted to that branch since it has no commits that aren't on develop yet.

Not sure why that happened, since I branched it off develop in my local machine :/ but I changed the base branch back to develop now.

isaisabel (changes requested):
Works great! Needs an entry in the changelog though.

isaisabel (approved):

Issue #120, opened 2019-11-22
Navigator version number in footer should be a link to the changelog


isaisabel:
As a user of the ATT&CK Navigator, I want to be able to easily access the changelog if I notice that the version number has incremented.

The version number displayed in the bottom left corner of the page should be made clickable such that it will open a new tab with the changelog.

Issue #178, opened 2020-06-23
Allow access to changelog in application


isaisabel:
As a user of the ATT&CK Navigator, I want to be able to see the changelog inside of the application, similarly to how the ATT&CK Website changelog is present [on the live site](https://attack.mitre.org/resources/changelog.html).

The changelog should be built from CHANGELOG.md so that there is a single source. A markdown renderer like [showdownjs](https://github.com/showdownjs/showdown) may be used to convert the markdown for display in the application.

isaisabel:
The documentation system for ATT&CK Workbench can be used as a basis for this feature.
- Markdown docs can be copied into the assets directory [using Angular](https://github.com/center-for-threat-informed-defense/attack-workbench-frontend/blob/develop/app/angular.json#L27).
- Markdown docs can be then be rendered in-app using [a markdown renderer](https://github.com/center-for-threat-informed-defense/attack-workbench-frontend/tree/develop/app/src/app/views/help-page).

PR #311, opened 2021-06-23
Added link to navigator version under the matrix

iguannalin:
Feature add for #120


isaisabel (changes requested):
I realize now that the phrasing of the issue was ambiguous, but this feature was intended to work alongside #178. 

Where the issue said "clickable such that it will open *a new tab* with the changelog" I meant a Navigator tab, not a browser tab. Back when the issue was opened the Navigator didn't use dialogs for docs (like the help page) but instead tabs like layers. 

The reason for wanting it in-app instead of a link is that the Navigator needs to be able to operate as fully as possible without an internet connection. To date I believe the only feature which doesn't work offline is the "view" links which reference attack.mitre.org. Everything else can be configured to run offline. That's also why the help page is available in-app instead of in a separate document (see also #179).

iguannalin:
This PR was closed because the issue was not fixed in this branch, it's been replaced by this PR: https://github.com/mitre-attack/attack-navigator/pull/323

PR #316, opened 2021-07-01
Feature/#179 help page

iguannalin:
Feature to add help page as external page from markdown


help.md
```diff
@@ -0,0 +1,901 @@
...
+  - Default Layers
+  - Disabling Features
+- Rendering Layers as SVG
+  - Measurement Units
+  - Configuring Image Size
+  - Configuring Text
+  - Customizing the Legend
+  - Display Settings
+- Exporting Layers to MS Excel
+- Notice
```
isaisabel, line 49: Without hyperlinks, this table of contents isn't particularly helpful. It appears as though it does have links in the rendered page, but [not but when viewed on GitHub](https://github.com/mitre-attack/attack-navigator/blob/feature/%23179-help-page/help.md#table-of-contents).
- On GitHub TOCs are automatically generated:
   image
- The TOC for the rendered markdown page could be inferred from the heading tags (see [this code example](https://github.com/center-for-threat-informed-defense/attack-workbench-frontend/blob/develop/app/src/app/views/help-page/help-page.component.ts#L30-L40)) and thereby auto-generated without needing to maintain a TOC in the markdown itself.

help.md
```diff
@@ -0,0 +1,901 @@
...
+    <li>
+        <p>
+            <b>Score Expression:</b> Technique
+            <span class="section-link linkName-Scoring_Techniques">scores</span> in the created layer are
+            initialized to the result of this expression. This field should take
+            the form of an equation or constant, using variables for layers.
+            Layer variables are shown in yellow on tabs when the "create layer
+            from other layers" panel is opened:
+        </p>
+        <img src="assets/image_scoreVariableExample.png" height="100px">
```
isaisabel, line 183: Rendered on GitHub the images do not work: 

- Shows as blank/broken image:
   image
- Or as their alt text if alt text is specified:
   image

This is because the images point to the assets folder but are missing part of the path for when opened on GitHub. The path is correct for when they're rendered in-app. When loaded in GitHub this is an example URL:
```
/mitre-attack/attack-navigator/raw/feature/%23179-help-page/assets/icons/ic_sort_alphabetically_ascending_black_24px.svg
```

The correct URL would be:
```
/mitre-attack/attack-navigator/raw/feature/%23179-help-page/nav-app/src/assets/icons/ic_sort_alphabetically_ascending_black_24px.svg
```
So it's missing the `/nav-app/src` part of the path for being rendered in GitHub.

In order to be displayed in both places properly you'll probably need to make the in-app version change the path to remove the `/nav-app/src` prefix which will be required for the GitHub view. I'm pretty sure u[pdating the renderer](https://github.com/jfcere/ngx-markdown#renderer0 for `img` tokens would be a workable approach for this.

help.md
```diff
@@ -0,0 +1,901 @@
...
+</p><br>
+<p>
+    Each layer created is independent of other layers. However, layers can be
+    <span class="section-link linkName-Creating_Layers_from_Other_Layers">combined</span>
+    in ways to support analysis, or
+    <span class="section-link linkName-Saving_and_Loading_Layers">saved locally</span>. Layer files are
+    saved in easy to parse and easy to generate JSON file so that ATT&CK data
+    can be used in other applications, analyzed beyond the capability of the
+    ATT&CK Navigator, and generated by tools for import into the Navigator.
+    The Layer file format is described <a href="assets/NavigatorLayerFileFormatv4_1.pdf">here</a>.
```
isaisabel, line 122: This hyperlink doesn't work when viewed in GitHub since the relative link to `assets` is incorrect when viewed in that context.

help.md
```diff
@@ -0,0 +1,901 @@
...
+    annotation of ATT&CK matrices, something that people are already doing
+    today in tools like Excel. We've designed it to be simple and generic -
+    you can use the Navigator to visualize your defensive coverage, your
+    red/blue team planning, the frequency of detected techniques or anything
+    else you want to do. The Navigator doesn't care - it just allows you to
+    manipulate the cells in the matrix (color coding, adding a comment,
+    assigning a numerical value, etc.). We thought having a simple tool that
+    everyone could use to visualize the matrix would help make it easy to
+    use ATT&CK.
+</p><br>
```
isaisabel, line 66: I realize it's a lot more work, but the inclusion of HTML tags for paragraphs is sort of un-markdown-y. HTML is valid markdown so it does work, but it's unnecessary to include and increases maintenance burden. Ideally these tags (`<p>` and `<br>`) should be removed altogether since an empty line between paragraphs is all that is actually required for paragraphs to render in Markdown.

help.md
isaisabel: Speaking of, it may be worth replacing the PDF version of the layer file format (which we always forget to update 😆 ) with a rendered copy of the actual document from the `/layers` folder.

nav-app/src/app/app-routing.module.ts
```diff
@@ -0,0 +1,18 @@
+import { NgModule } from '@angular/core';
+import { RouterModule, Routes } from '@angular/router';
+import { HelpPageComponent } from './help-page/help-page.component';
+import { TabsComponent } from './tabs/tabs.component';
+
+const routes: Routes = [
+  { path: '', component: TabsComponent },
+  { path: 'help', component: HelpPageComponent }
```
isaisabel, line 8: While this does work, I'm curious as to what it achieves? It doesn't seem like the `/help` route is linked to from anywhere else in the app. It does allow one to link directly to `/help` on the internet, but you can already link to the help markdown file on GitHub for the same effect (aforementioned bugs with that markdown notwithstanding).

isaisabel: Or was my assumption that the markdown was intended to be consumed in GitHub itself incorrect, and the intention was to link to `/help` from the beginning? 

That would explain why it was so buggy when viewed in Markdown. But since it lives in the root directory of the repo I expect people would look at it there. We may want to move it somewhere within `src` (e.g the `assets` folder) if it's not supposed to be consumed directly on GitHub.

isaisabel: Consuming it in Markdown on GitHub was how [the issue originally described it](https://github.com/mitre-attack/attack-navigator/issues/179#issue-643814179), but we can definitely change that if it's not easy to achieve due to the aforementioned image issues.

nav-app/src/app/datatable/data-table.component.html
```diff
@@ -556,15 +556,15 @@
...
+    <div class="noselect" style="padding:4px; font-size:14px; position: absolute; width: 100%; text-align:center">
         legend
     </div>
 </div>
 
 <div class="legend" *ngIf="showingLegend && configService.getFeature('legend')">
     <div class="legendBar" (click)="showingLegend = !showingLegend" style="position: absolute; top:0px; left:0px;">
         <img src="assets/icons/ic_keyboard_arrow_down_black_24px.svg" style="position: absolute; left:4px; top: 3px;">
-        <div class="noselect" style="padding:4px; font-size:14px; position: absolue; width: 100%; text-align:center">
+        <div class="noselect" style="padding:4px; font-size:14px; position: absolute; width: 100%; text-align:center">
```
isaisabel, line 567: These changes _may_ be the cause of the legend issues mentioned previously:

image

nav-app/src/app/datatable/data-table.component.scss
```diff
@@ -368,6 +368,7 @@ $panel-light: lighten($panel-dark, 8%);
 }
 
 .legend {
+    z-index: 999;
```
isaisabel, line 371: Again a possible cause of the legend bug

isaisabel (changes requested):
I think there may be some differing views on requirements/approach here. It seems like the approach taken doesn't leave the markdown file usable on GitHub directly despite its inclusion in the root of the repo. It seems like the way it was developed may have been intending to use routing (`/help`) to host the docs instead of putting them on the repo itself. But if that was the design, why is `help.md` included in the repo root at all?

Additionally:
- The help markdown file (if intended to be accessed by users directly on GitHub for documentation) should be called `USAGE.md` for consistency with other ATT&CK Software projects.
- The help dialog seems to have a strange interaction with the legend when the help page is opened when on a layer tab (that white box is apparently the legend): 
   image
- The main repo documentation (README.md) should mention the availability of usage documentation and link to it, however it is hosted. Probably at the end of the first paragraph of the Usage section would be prime placement for a link to the document.

isaisabel:
To be clear I'm not saying that the approach of hosting it via the Angular router on `/help` is wrong, just that it seems a bit inconsistent in terms of how this was actually implemented. It seems like part of both approaches (using `/help` and having it available on GitHub directly) was implemented.

USAGE.md
```diff
@@ -0,0 +1,561 @@
...
+Tactic-spanning Techniques are evaluated individually: if a technique is annotated differently in two tactics, the
+output layers' techniques will honor this difference.
+
+<b>Tip:</b> Score expressions don't need to use variables! You can use this to create a new layer with a constant score
+for each technique. For example, if you wanted a new layer where all techniques are scored 50, you could simply type 50
+into the score expression input.
+
+# Layer Controls
+
+## ![Layer Information](nav-app/src/assets/icons/ic_description_black_24px.svg) Layer Information
```
iguannalin, line 122: Only small issue left to fix here is that in GH this heading anchor gets generated as 'user-content--layer-information' (because of the icon image markdown), which causes all the anchor links in other places back to this header not work correctly

USAGE.md
```diff
@@ -0,0 +1,561 @@
+# MITRE ATT&CK® Navigator Version 4.4
```
isaisabel, line 1: We should probably remove the version numbers here. Previously we included them because they could inserted by Angular, but since they can't in the Markdown they're a likely thing to forget to update when releasing new versions of the application.

USAGE.md
```diff
@@ -0,0 +1,561 @@
...
+Bug reports and feature requests can be submitted to
+our <a href="https://github.com/mitre-attack/attack-navigator/issues">GitHub Issue Tracker</a>. The source code for the
+ATT&CK Navigator can be retrieved from our <a href="https://github.com/mitre-attack/attack-navigator">GitHub
+repository</a>.
+
+<!-- sections -->
+
+# Layers
+
+A layer constitutes a view of the tactics and techniques matrix for a specific technology domain. Version 4.4 of the
```
isaisabel, line 24: Another place where the version number should be removed.

USAGE.md
isaisabel: This is still an observed issue as far as I can tell?

USAGE.md
```diff
@@ -0,0 +1,561 @@
...
+    - <code>a+b+c</code>
+    - <code>a*(b+c)</code>
+    - <code>100-a</code> (reverse the scoring of <code>a</code>, assuming <code>a</code> is on a 0-100 scale)
+    - <code>(a>b) and (a>=75)</code>
+
+- <b>Gradient:</b> choose which layer to inherit <a href="#scoring-gradient">scoring
+  gradient</a> from.
+- <b>Coloring:</b> choose which layer to inherit <a href="#assigning-manual-colors">manually
+  assigned technique colors</a>
+- <b>Comments:</b> choose which layer to inherit <a href="#adding-comments-to-techniques">
```
isaisabel, line 105: This link for and others like it seem to be broken in the GH view, the actual fragment is `#-adding-comments-to-techniques`. Probably because there is an image in the header ([related comment](https://github.com/mitre-attack/attack-navigator/pull/316/files/cd270a767ce3b0839a80c5fec218ff32fca47d08..9ff30b8cc50e792b6f7f9aabac34a220fe6fb4b5#r665077245))?

isaisabel (changes requested):
This seems to have fixed the vast majority of the bugs. There is still a small issue with some of the relative links when viewed on GitHub. Also, we need to make mention of the USAGE document in the main repo README (the following can be appended to the 5th paragraph under the Usage section)
> Please see Install and Run for information on how to get the ATT&CK Navigator set up locally, and the USAGE document for more information about how to use the application.

Aside from that I don't really have any concerns with these changes. 

One minor improvement we may want to consider is reducing the width of the help dialog, as [readability decreases with long line lengths](https://practicaltypography.com/line-length.html). 35em is roughly correct to Butterick's recommendation of 45-90 characters, but the actual dialog may need to be slightly wider to account for text margins. Dialog sizes can be set where the dialog is opened ([e.x](https://github.com/center-for-threat-informed-defense/attack-workbench-frontend/blob/develop/app/src/app/views/stix/stix-page/stix-page.component.ts#L139-L140)).

help.md
iguannalin: I think this is the only thing left in the story I haven't touched on yet. I can replace this link to the markdown version in the [/layers](https://github.com/mitre-attack/attack-navigator/blob/master/layers/LAYERFORMATv1.md) folder instead. Should I have it rendered in-app as well?

isaisabel: Yes, I think it would be best as rendered in-app. Since the help page is in a dialog I'm not sure the best way to do that (close the original dialog and open the new one? Replace the dialog contents?) since spawning a dialog on top of a dialog is bad UX.

We'll still need to update the settings for which markdown file to render when we update the layer file, unless we can do something fancy with the layer version in globals.ts to infer which file to load. But at least we won't have to maintain a PDF version of the layer spec anymore.

iguannalin: @isaisabel I added a new component to dynamically display the correct layer information markdown, based on the global nav ver: https://github.com/mitre-attack/attack-navigator/tree/feature/%23179-help-page/nav-app/src/app/layer-information

Let me know what you think!

nav-app/src/app/layer-information/layer-information.component.scss
```diff
@@ -0,0 +1,36 @@
...
+
+    table {
+      td {
+        border-top: 1px solid $panel-dark;
+      }
+
+      td:not(td:last-child) {
+        border-right: 1px solid $panel-dark;
+      }
+    }
```
isaisabel, line 30: Should add border-collapse to greatly improve the table aesthetics:
### Before
Screen Shot 2021-07-13 at 10 00 33 AM

### After
Screen Shot 2021-07-13 at 10 00 39 AM

isaisabel: Also increasing padding:
Screen Shot 2021-07-13 at 10 03 15 AM

isaisabel: And (possibly?) removing the column dividers.

Screen Shot 2021-07-13 at 10 03 49 AM

isaisabel: Note that padding needs to be applied to both `th` and `td`, I forgot to apply it to `th` in my above screenshots.

See [Butterick's Practical Typography](https://practicaltypography.com/tables.html) for some tips and tricks for making aesthetic tables

nav-app/src/app/help/help.component.scss
```diff
@@ -1,47 +1,54 @@
...
+  }
 
-// table of contents
-.toc {
-    list-style: none;
-    ul {
-        list-style: none;
+  .content {
+    overflow-y: scroll;
+    max-height: 60vh;
```
isaisabel, line 32: Adding `padding-right: 25px;` to the content class helps separate the scrollbar from the text, increasing readability.

### Before
Screen Shot 2021-07-13 at 10 10 02 AM

### After
Screen Shot 2021-07-13 at 10 10 15 AM

Removing the gutter of the scrollbar may also improve the UI (though it's not always easy to do that in all browsers). [Here's one way to do that](https://github.com/center-for-threat-informed-defense/attack-workbench-frontend/blob/develop/app/src/style/theme.scss#L7-L15).

isaisabel: This padding also needs to get propagated to the layer format document UI -- the tables there are especially strange looking without the padding.

It may be worth extracting these rendered-markdown styles into an abstract class (`content { @extends %markdown-common; }` or common class (`.rendered-markdown` in the template) in a separate file to reduce code reuse.

isaisabel (changes requested):
There are a few places for improvement with the new layer document features and some of the formatting but it's definitely getting there!

nav-app/src/app/help/help.component.scss
isaisabel: Note: scrollbar gutters should ideally only be removed for these dialogs, not across the application. I think the Layer box itself would look weird without the gutters.

USAGE.md
```diff
@@ -43,7 +43,7 @@ Each layer created is independent of other layers. However, layers can be
 <a href="#saving-and-loading-layers">saved locally</a>. Layer files are saved in easy to parse and easy to generate JSON
 file so that ATT&CK data can be used in other applications, analyzed beyond the capability of the ATT&CK Navigator, and
 generated by tools for import into the Navigator. The Layer file format is
-described <a href="nav-app/src/assets/NavigatorLayerFileFormatv4_1.pdf">here</a>.
+described <a href="layer-information-dialog">here</a>.
```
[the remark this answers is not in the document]
iguannalin: Good catch! Thanks

USAGE.md
```diff
@@ -43,7 +43,7 @@ Each layer created is independent of other layers. However, layers can be
 <a href="#saving-and-loading-layers">saved locally</a>. Layer files are saved in easy to parse and easy to generate JSON
 file so that ATT&CK data can be used in other applications, analyzed beyond the capability of the ATT&CK Navigator, and
 generated by tools for import into the Navigator. The Layer file format is
-described <a href="layer-information-dialog">here</a>.
+described <a href="layers/LAYERFORMATv4_2.md">here</a>.
```
isaisabel, line 46: Maybe this should just be the `/layers` folder instead of the specific file? Otherwise we will have to update the hyperlink when we update the layer format. It seems like that should work with your link replacement since it only cares about the `layers/` part.

isaisabel (approved):
My only comment is to update the hyperlink in the markdown to be less specific to the particular layer. I'll push that myself since it's a 1 LOC change and then merge the PR.

PR #318, opened 2021-07-09
Layer Upgrade UI

### LINKED ISSUE #181: Add "Layer Upgrade" UI
As a user of ATT&CK, I want to be able to upgrade layers created in older versions of ATT&CK with an intuitive UI.

- This UI should perform the same operations as the [update-layers script](https://github.com/mitre-attack/attack-navigator/tree/develop/layers#update-layerspy) automatically, and provide notification to the users of the revocations that were followed.
- The UI should notify the user of what objects have changed between the two versions (see also [diff_stix.py](https://github.com/mitre-attack/attack-scripts/blob/master/scripts/diff_stix.py) for other implementations showing changes between STIX bundles). 
    - It should highlight changes in two lists, one for objects which have changed which had annotations, and one for objects which changed which did not have annotations. These lists should be shown as adjacent columns or tabs.
    - Changes should be grouped by type: 
        - Changes which removed the object (deletion or revocation) and have therefore removed annotations the user had added.
        - Changes to the object content (major and minor changes), which may require updated annotations
        - Object additions, which may require new annotations
        - which objects have been upgraded automatically
    - Possibly allow the user to filter the lists to show only techniques, or all objects (which would be useful if the user was tracking mitigations or APT coverage)

---

This UI should take the form of a sidebar adjacent to the Matrix. Clicking on a change or section should select the object/objects within the section where they occur in the matrix, and likewise hovering should highlight objects where they occur in the matrix. Highlights and selections should also work the other direction (where they occur in the matrix they also occur in the sidebar). 

For objects which changed that have annotations (in particular cases where the object is no longer in the matrix, e.g revocations and deletions), the previous annotations should be shown on the item so that the user can remap them.

The user should be able to minimize and expand the sidebar if they so choose.

clemiller:
Implements the Layer Upgrade workflow in the sidebar to upgrade layers created in older versions of ATT&CK.
- Displays technique changes that occurred between the two versions
- Allows users to map annotations from the previous version to the current version
- Techniques with tactics that do not map one-to-one between versions have a drag & drop interface
- Each step can be filtered to show only techniques which have annotations
- Provides links to attack.mitre.org for both versions of each technique

Resolves #181


nav-app/src/app/layer-upgrade/layer-upgrade.component.html
```diff
@@ -0,0 +1,191 @@
...
+                <input id="select_{{section}}" class="checkbox-custom" type="checkbox"
+                    [checked]="allReviewed(section)" (click)="reviewAllChanged(section)">
+                <label for="select_{{section}}" class="checkbox-custom-label noselect">select all</label>
+            </div>
+
+            <!-- techniques -->
+            <ng-container *ngFor="let attackID of changelog[section]">
+                <mat-expansion-panel *ngIf="!showAnnotatedOnly || anyAnnotated(attackID)">
+                    <mat-expansion-panel-header>
+                        <mat-panel-title>
```
isaisabel, line 47: The content of the header doesn't seem to be center aligned. Some elements are above or below the midline and it's a bit inconsistent between elements as well:
Screen Shot 2021-07-12 at 1 04 20 PM

nav-app/src/app/layer-upgrade/layer-upgrade.component.html
```diff
@@ -0,0 +1,191 @@
...
+
+        </mat-expansion-panel>
+        </mat-accordion>
+    </div>
+
+    <mat-divider></mat-divider>
+
+    <!-- done button -->
+    <div class="done-button">
+        <span>Updated {{changelog.reviewed.size}}/{{changelog.length()}} techniques</span>
```
isaisabel, line 171: Should the total count here take into account the "show annotated techniques only" count? Otherwise it implies a lot of work has been skipped for users who only want to adjust existing annotations and not review literally every change type for unannotated techniques.

nav-app/src/app/layer-upgrade/layer-upgrade.component.html
```diff
@@ -0,0 +1,191 @@
...
+                </mat-expansion-panel>
+            </ng-container>
+
+            </ng-template> <!-- end changelog lazy render -->
+
+        </mat-expansion-panel>
+        </mat-accordion>
+    </div>
+
+    <mat-divider></mat-divider>
```
isaisabel, line 167: There is uneven spacing around this divider:
image

isaisabel (changes requested):
- UI looks great overall! Very clean and consistent with the other sidebar tool.
- The upgrade prompt dialog that asks users if they want to upgrade should probably mention that there is a workflow so people know what's going to happen. Currently it seems unchanged from previously and users may opt not to upgrade if they don't know there's now a tool for it.
- Select techniques across tactics doesn't seem to work properly when selecting in the sidebar. At least for highlighting it seems to highlight across tactics, but the selection does work properly as far as I can tell.
- I like how "existing" annotations in the layer do show (e.g if the user added an annotation on an object it does show up in the "new" column of the upgrade UI even before the copy)
- Cells in the "old" column appear as if they are selectable -- they have highlight and cursor:pointer -- but nothing happens when they are themselves clicked. Is this a bug? If no behavior is supposed to occur when clicked they shouldn't be interactive in this way.
- Bug: if you select a cell in the matrix and unselect it in the sidebar, you can keep the annotation dialogs open even when nothing is selected:

https://user-images.githubusercontent.com/28868888/125321571-d3f29280-e30a-11eb-9e5c-c2c636025db2.mov

- Needs help page documentation (may want to wait until the behavior is finished and for #316 to be completed though -- I'm guessing this was an intentional omission for this early draft)
- In the additions section objects should only appear in the rightmost column since they're not in the left (old) version of ATT&CK: 

   image
- It seems a bit weird that "additions" aren't shown unless the user enables "show annotated techniques only." It makes sense, but from a UX perspective I would think users want to:
    - Review changes to annotated techniques and adjust annotations appropriately
    - Know what objects were added (which needs coverage)
   
   But the second use case isn't easy to get to unless you want to see _everything_. Not sure if there's an easy solution without redesigning the entire "show annotated techniques only" approach, but it's a bit unfortunate additions are so hard to get to without having to go into the more comprehensive mode.
- A possible adjustment to address this UX issue would be to increase the number of options for display instead of just having the single option. 
    - Show new techniques
       - Shows/hides the additions section 
    - Show technique changes without annotations
       - Affects contents of changes, minor changes, deprecations and revocations sections
    - Show unchanged techniques
       - Shows/Hides the unchanged section
   Note that this wouldn't hide the sections themselves, just the relevant contents. They'd still show up in the section list but just be disabled if the relevant button was disabled.

nav-app/src/app/layer-upgrade/layer-upgrade.component.html
```diff
@@ -0,0 +1,191 @@
...
+                            <mat-icon class="info" matTooltipPosition="above" 
+                                matTooltip='To copy annotations, drag a cell with annotations from ATT&CK {{compareTo.version}} to a cell in ATT&CK {{viewModel.version}}.'>
+                                info_outline
+                            </mat-icon>
+                        </div>
+
+                        <!-- technique from latest version-->
+                        <div class="float wide" *ngIf="getTechnique(attackID, viewModel, section) as newTechnique">
+                            <div class="changelog-cells" *ngFor="let tactic of getTactics(attackID, viewModel, section)"
+                                dndDropzone (dndDrop)="onDrop($event, attackID, newTechnique, tactic)">
```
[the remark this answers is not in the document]
isaisabel: There's actually a console error when you drop to another technique, so this might take some actual implementation to avoid errors even if the UI can't be fixed:

Screen Shot 2021-07-12 at 1 07 17 PM
.

USAGE.md
```diff
@@ -64,7 +64,42 @@ layer description, view configuration.
...
+following change types:
+- <b>Additions:</b> new techniques that have been added since the previous version
+- <b>Changes:</b> changes to techniques
+- <b>Minor Changes:</b> changes to techniques without version increments
+- <b>Deprecations:</b> techniques which have been deprecated
+- <b>Revocations:</b> techniques which have been revoked by a new technique
+- <b>Unchanged:</b> techniques with no changes between the two versions
+
+By default, only techniques which were previously annotated are shown in the interface. The filters at the top of the interface
+can be used to individually adjust which change types are shown.
```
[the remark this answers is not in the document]
isaisabel: In the filters section, the use case for each filter should be specified. E.g:
- show additions: shows techniques that were added since the version of ATT&CK the layer was created for. Enable this if you want to review newly added techniques to determine if they warrant annotation.

USAGE.md
```diff
@@ -64,7 +64,42 @@ layer description, view configuration.
...
 configuration file. Doing so will restore your saved layer to the ATT&CK Navigator. This interface also has a "load from
-URL"input allowing you to open a layer json from a remote source.
+URL" input allowing you to open a layer json from a remote source.
+
+## Upgrading a Layer to the Current Version
+
+A layer uploaded with an outdated version of ATT&CK will prompt you to upgrade to the current version. Choosing "yes" will
+initiate the layer upgrade workflow in the Navigator sidebar. A new, blank layer in the current ATT&CK version and the layer's
+domain will be loaded in a new tab; previous technique annotations can be found in the interface and mapped to techniques in
+the current version, as described below.
```
isaisabel, line 74: This writing accurately describes what the layer upgrade interface is, but not how it is intended to be used (what the "workflow" actually is). That context is really important to understanding why the different features of the UI are useful. 

> This interface allows users to upgrade an ATT&CK Navigator layer created on a previous version of ATT&CK to the current version of the dataset. The interface allows users to step through each change in the set of techniques in the dataset:
> - The user can see what techniques have been added to the dataset since the layer was created, and annotate them as necessary.
> - The user can see whether the techniques that they had previously annotated have changed, and adjust the annotations accordingly.
> - The user can determine if any annotated techniques have been removed or replaced by new techniques, and in the latter case copy annotations to the replacing technique(s).
> - The user can also verify what techniques haven't been changed since the layer has been created.

USAGE.md
```diff
@@ -64,7 +64,42 @@ layer description, view configuration.
...
+- <b>Deprecations:</b> techniques which have been deprecated
+- <b>Revocations:</b> techniques which have been revoked by a new technique
+- <b>Unchanged:</b> techniques with no changes between the two versions
+
+By default, only techniques which were previously annotated are shown in the interface. The filters at the top of the interface
+can be used to individually adjust which change types are shown.
+
+### Mapping Annotations
+
+Each technique section contains cells for each tactic it is found under in the previous version (with any annotations that
```
isaisabel, line 90: > cells

I'm not sure if this word is intuitive here. Generally this first sentence is a bit hard to understand if you've never see the UI before, but I can't think of much that can be done to improve that.

USAGE.md
```diff
@@ -64,7 +64,42 @@ layer description, view configuration.
...
+were previously placed on that object) and in the current version. There are two approaches to copying annotations from the
```
isaisabel, line 91: > There are two approaches to copying annotations from the previous layer:

This is the first mention of copying annotations in this section. I think the section really just needs an introductory sentence stating what it's _about_:

> For annotated techniques which have changed since the layer was created, you may want to update the set of annotations to account for changes to scope or content in the technique in question. 
>
>Each technique in the layer upgrade interface lists its previous and current state, with links to [attack.mitre.org](https://attack.mitre.org/) for both versions of the technique to enable easy review. You can then copy annotations from the previous version of the technique to the current one, making adjustments if nessecary.
> [insert the stuff about techniques whose tactics have changed here]

USAGE.md
```diff
@@ -64,7 +64,42 @@ layer description, view configuration.
...
+- For techniques with tactics that map one-to-one between the two versions, annotations can be copied by clicking the ">"
+button next to the tactic. This will copy any existing annotations on the tactic to the technique under the same tactic in
+the current version.
+- For techniques with tactics that have changed between the two versions, annotations can be copied by dragging & dropping the
+tactic cell from the previous version to the tactic(s) in the current version.
+
+Techniques can be marked as reviewed under their respective expansion panels to visually indicate that you have reviewed that
+object. Additionally, the "mark all as reviewed" button will mark all of the technique objects currently shown under the change
+type section as reviewed. Once the annotations have been remapped to the current version, you can click the done button to close
+the sidebar; you will not be able to return to the layer upgrade interface.
```
isaisabel, line 102: > Once the annotations have been remapped to the current version, you can click the done button to close
the sidebar; you will not be able to return to the layer upgrade interface.

- Should acknowledge that not all annotations _have_ to be mapped for all use cases:
> Once you're done upgrading the layer, you can click the...
- Should mention the updated fraction count:
> The "updated X/Y Techniques" display next to the done button can be used to determine if you missed any techniques. The numerator of the fraction reflects the number of techniques you have marked as "reviewed". The denominator denotes the total number of techniques according to your configuration (for example, if you had disabled "count additions" they would not be counted against this total).

USAGE.md
```diff
@@ -64,7 +64,42 @@ layer description, view configuration.
 Saved layer configuration files can be opened in the ATT&CK Navigator to restore a layer you've worked on previously. To
```
isaisabel, line 64: This is surely a merge conflict, but the table of contents in the help page is... pretty wild. For some reason it has the ".narrow" class which makes the width 20px:

image

nav-app/src/app/layer-upgrade/layer-upgrade.component.html
```diff
@@ -21,142 +32,147 @@ <h2>Layer Upgrade</h2>
...
-                        <div class="float arrow"><mat-icon inline="true">east</mat-icon></div>
-                        <div class="float wide">
-                            <div class="version">ATT&CK {{viewModel.version}}</div>
-                            <mat-divider></mat-divider>
+            <!-- techniques -->
+            <mat-expansion-panel *ngFor="let attackID of applyFilters()[section]" #panel="matExpansionPanel">
+                <mat-expansion-panel-header>
+                    <mat-panel-title>
+                            <mat-icon *ngIf="isReviewed(attackID)" class="reviewed">check_circle_outline</mat-icon>
+                            <mat-icon *ngIf="!isReviewed(attackID)" class="disabled">radio_button_unchecked</mat-icon>
```
isaisabel, line 60: The unchecked state here still appears as if it is clickable. It really needs a deemphasis color (lighter gray) to show that it's not supposed to be interacted with. See also: https://github.com/mitre-attack/attack-navigator/pull/318#discussion_r668050717

isaisabel (commented):
In summary, I think this is really good and I like all the improvements that were made. I do think there is room for improvement though. The current implementation (and documentation) doesn't really make it obvious what the "workflow" actually _is_, even though it does support the workflow. The proposed improvements below _should_ help guide users into what they're supposed to be doing with the interface.

---

The set of filters and their behavior is I think correct to match the use cases for this feature. However, the way they behave is not intuitive -- I think this is a problem with the documentation and naming of the filters and not the actual filter implementation itself.

- "Show annotated techniques only" and "show technique changes" should be mutually exclusive? Maybe? It's basically toggling between showing all techniques that have changes or only the changed techniques that have annotations. 
- The interaction between "show annotated techniques only" and the other filters doesn't behave logically. In the implementation "show annotated techniques only" doesn't affect additions (if that other filter is enabled), but the name of the filter implies it would.
- The word "show" seems... off somehow. It's accurate, but what we're really doing here (from a workflow standpoint) is determining which techniques are considered necessary to update to complete the workflow. "show" doesn't convey this.

Here's a mockup of a revised UI that may better convey the behavior of the filters:

example

- Note: example is missing minor changes subsection and unchanged section
- Each section can be enabled/disabled separately
- Minor changes can be disabled independently of major changes
- Show annotated only now obviously only applies to changes
    - Maybe rename to "show only previously annotated"? Is that clearer? 
- The (?) marking should provide a tooltip explaining the purpose/behavior of that control. In this case it would explain that minor changes do not include changes to technique scope and are typically only typo corrections. We may want to add similar info to all of the filter controls.
- Maybe "count" in this mockup could be "review"? That _might_ state the intention better?

Also, the total updated count needs to reflect the current set of filters. For example:
image

The correct fraction should in this case be 0/8 since there are only 8 techniques shown with the current set of filters.

The updated fraction also uses the term "updated" where elsewhere it is "reviewed". This terminology should probably be consolidated -- IMO in favor of "Reviewed" since sometimes reviewing a technique change might not result in an update to the annotation.

Lastly, changing the set of filters causes quite a bit of UI lag while the new set of techniques is loaded (or the UI updates, not quite sure). There should be a loading marker while the interface updates to ensure that the user doesn't interpret the loading as a crash. The `setTimeout` trick to force a render cycle can be used to render a spinner overlay before the actual loading begins (which would delay the render cycle otherwise).

isaisabel (changes requested):
Forgot to "request changes" in the review I just submitted, whoops!

USAGE.md
```diff
@@ -68,38 +68,51 @@ URL" input allowing you to open a layer json from a remote source.
...
+- <b>Unchanged:</b> Techniques which have not changed between the two versions.
 
 ### Mapping Annotations
 
-Each technique section contains cells for each tactic it is found under in the previous version (with any annotations that
-were previously placed on that object) and in the current version. There are two approaches to copying annotations from the
-previous layer:
+For annotated techniques which have changed since the layer was created, you may want to update the set of annotations to account 
+for changes to scope or content. Each technique in the interface lists its previous and current state, with links to 
+<a href="https://attack.mitre.org/">ATT&CK&reg;</a> for both versions of the technique to enable easy review. In steps with techniques you
```
isaisabel, line 93: ATT&CK hyperlink here should be "the ATT&CK Website" instead of just "ATT&CK." ATT&CK is the knowledge base, the ATT&CK Website is the tool used to interface with it.

Also the registered trademark is only required for the first occurrence of ATT&CK on a given page. In this case the header at the top of the help page covers the trademark.

USAGE.md
```diff
@@ -68,38 +68,51 @@ URL" input allowing you to open a layer json from a remote source.
...
+
+### Change Types
+
+The workflow steps through the changes to the dataset that occurred between the two versions, grouped by the following change types:
+- <b>Additions:</b> New techniques that have been added to the knowledge base.
+- <b>Changes:</b> Techniques with major updates such as changes to scope, new reporting, and so forth.
+- <b>Minor Changes:</b> Techniques with minor updates such as typo corrections.
+- <b>Revocations:</b> Techniques which have been replaced by other techniques.
+- <b>Deprecations:</b> Techniques which have been removed from the dataset.
+- <b>Unchanged:</b> Techniques which have not changed between the two versions.
```
isaisabel, line 87: These two sections seem duplicative. I think you can probably merge them so that the prior section entries (the use cases) appear under the change types. E.g:
- **Additions:** New techniques that have been added to the knowledge base. View techniques which have been added to the dataset since the layer was created and annotate them as necessary.
- **Changes:** Techniques with major updates such as changes to scope, new reporting, and so forth. Review updated techniques and adjust annotations accordingly.

USAGE.md
```diff
@@ -68,38 +68,51 @@ URL" input allowing you to open a layer json from a remote source.
...
+Techniques can be marked as reviewed under their respective panels to visually indicate that they have been reviewed. The "reviewed 
+X/Y techniques" displayed underneath the techniques list can be used to determine if you missed any techniques. The numerator reflects 
+the number of techniques you have marked as "reviewed" and the denominator denotes the total number of techniques shown according
+to your configuration (for example, if you have enabled the "show annotated techniques only" filter, only the number of techniques
+which are annotated are counted in this total).
+
+You can view and verify the status of the layer upgrade at the end of the workflow to ensure you have created or adjusted all annotations 
+as desired. A status is displayed next to each section name to indicate either the number of techniques you have reviewed in the section, 
+if you have skipped the section, or if there are no techniques to review in the section. Once you have finished upgrading the layer, you 
+can click the "done" button to close the sidebar.
```
isaisabel, line 113: This is really well written, bravo 👏 !

isaisabel (changes requested):
This is a fantastic improvement and I really only have minor comments at this point. There is really only one behavioral change, the rest is just documentation improvements.

isaisabel (changes requested):
The changes all look good here, I have no other comments on the layer upgrade UI. 

I have discovered one comparability issue with the current development branch: the "search & multiselect merge notification" added in #336 could appear alongside the layer upgrade sidebar if the first thing a user does when they open the v4.4 Navigator is upgrade a layer.

image

The merge notification popover should not be shown if the user is upgrading the layer. Adding an extra check to this statement to determine if the layer upgrade dialog is present should produce the correct behavior: https://github.com/mitre-attack/attack-navigator/blob/develop/nav-app/src/app/search-and-multiselect/search-popover-notification/search-popover-notification.component.ts#L79

With that comparability issue resolved this PR is suitable to merge.

isaisabel (approved):

PR #322, opened 2021-07-16
Fix/#204 technique hover

iguannalin:
Fixed bug where the `select techniques across parents` / `select sub-techniques with parent behavior` on hover were not correct.

Only concern I have is that I've removed the [line comparing the attackIDs of the current and highlighted techniques to see if it should be highlighted](https://github.com/mitre-attack/attack-navigator/blob/a7d4703365db515c472d0c06436227c83eb26015/nav-app/src/app/matrix/technique-cell/technique-cell.component.ts#L38):
`if (compare.attackID == compareTo.attackID) return true;`
This was removed because instead of storing Technique in the highlightedTechniques, I'm now saving an array of the technique IDs, for faster lookup (and because the search component needs it as an array). I'm hoping that checking both the technique and tactic when the user hovers over a cell in the matrix should suffice.

### DIFF
--- a/nav-app/src/app/matrix/technique-cell/technique-cell.component.ts
+++ b/nav-app/src/app/matrix/technique-cell/technique-cell.component.ts
@@ -24,33 +24,22 @@ export class TechniqueCellComponent implements OnInit {
         if (this.showContextmenu) return false;
         if (this.viewModel.highlightedTechniques.size === 0) return false;
 
-        return (this.viewModel.highlightedTechniques.has(this.technique.id) && this.viewModel.highlightedTactic && this.viewModel.highlightedTactic.id == this.tactic.id);
+        return (this.viewModel.highlightedTechnique === this.technique && this.viewModel.highlightedTactic && this.viewModel.highlightedTactic.id === this.tactic.id);
     }
 
-    // isHighlighted() checks for three cases:
-    //     1) if selectTechniquesAcrossTactics is enabled,
-    //     2) if selectSubtechniquesWithParent is enabled,
-    //     3) if highlightedTactic exists
-    // If 1) is enabled, it searches through the highlightedTechniques array for the current technique's id
-    // And if 2) is enabled, it searches through the highlightedTechniques array for the current technique's parent's id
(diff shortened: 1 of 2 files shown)


isaisabel (changes requested):
This fix does seem to fix the highlight behavior with "select sub-techniques with parent." However, there is another bug observed in that feature which was not fixed (but was probably introduced at the same time).

The tooltip for the parent also shows when the child is hovered, which isn't intended behavior:

image

Compare this to the behavior on master, where the tooltip for the parent isn't shown:

image

nav-app/src/app/matrix/technique-cell/technique-cell.component.ts
```diff
@@ -24,25 +24,25 @@ export class TechniqueCellComponent implements OnInit {
...
-            if (this.viewModel.selectSubtechniquesWithParent) {
-                return this.technique.isSubtechnique && this.viewModel.highlightedTechniques.has(this.technique.parent.id);
+        let isHighlighted = this.showContextmenu;
+        let idToMatch = this.technique.id;
+        if (this.viewModel.selectSubtechniquesWithParent && this.technique.isSubtechnique) idToMatch = this.technique.parent.id;
+
+        if (this.viewModel.highlightedTechniques.has(idToMatch)) {
+            if (!this.viewModel.highlightedTactic) { // highlight is called from search component
+                return true;
+            } else if (this.viewModel.highlightedTactic) {
```
isaisabel, line 39: Is there any functional difference between the `else if` used here and just an `else`? 
```
if (!this.viewModel.highlightedTactic) [...]
else if (this.viewModel.highlightedTactic)
```
Is the same as:
```
if (!this.viewModel.highlightedTactic) [...]
else
```
Because the `else if` is the inversion of the `if`.

isaisabel (approved):
Seems to fix the bug and definitely simplifies the isHighlighted code into a more understandable implementation. There is one case of an else if seeming unnecessary but it's fine to merge regardless.

isaisabel (approved):

PR #329, opened 2021-08-04
Added check for if Safari version is 13. and below to show incompatibility warning dialog

iguannalin:
Fix uses the existing `is_js `dependency.

This fix was tested using the Safari browser's Develop tool (via user agent in the menu) to run the app using version 13.0 and 14.0 on my Mac. It doesn't seem possible to downgrade Safari versions, so this seems to be the best way of testing it.


isaisabel (approved):
Unfortunately I was unable to dig up a computer with this version of safari on it to test, but it looks good to me. I also pushed a minor documentation update to clarify the safari support.

PR #337, opened 2021-08-12
Disable Comment Underline

### LINKED ISSUE #268: Allow user to hide the comment underline
As a user, I want to be able to hide the comment underline. This would be useful to declutter the UI for layers where all techniques have comments.

clemiller:
Allows users to disable the comment underline effect on techniques.

Resolves #268

### DIFF
--- a/nav-app/src/assets/config.json
+++ b/nav-app/src/assets/config.json
@@ -98,7 +98,6 @@
     },
 
     "comment_color": "yellow",
-    "show_comment_underline": true,
 
     "banner": "",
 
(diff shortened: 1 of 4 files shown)


nav-app/src/assets/config.json
```diff
@@ -98,6 +98,7 @@
     },
 
     "comment_color": "yellow",
+    "show_comment_underline": true,
```
isaisabel, line 101: I think this should probably be a property of the `features` array. That way it can be customized in the _create a customized Navigator_ interface. Otherwise you could already get this behavior by setting `comment_color` to `"transparent"`.

isaisabel (changes requested):
Works fine as implemented, but there's a low-hanging-fruit improvement to be made in terms of where the configuration is stored.

isaisabel (approved):
Perfect!

PR #339, opened 2021-08-12
Added alert and throw console error when external id is not found in imported STIX data

iguannalin:
Simple alert for external_id not found error

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ Version 4.4 of the Navigator restores Safari support provided you are using Safa
 - Improved favicon for standardization with other ATT&CK tools.
 - Show technique even if it's disabled, as long as any of its sub-techniques are enabled. See issue [#298](https://github.com/mitre-attack/attack-navigator/issues/298)
 - Converted help menu to USAGE.md that can be accessed as a raw markdown outside of app. See issue [#179](https://github.com/mitre-attack/attack-navigator/issues/179).
-- Application now alerts the user when input STIX data contains an object missing an ATT&CK ID instead of failing silently. See issue [#178](https://github.com/mitre-attack/attack-navigator/issues/178).
+- Added dialog for CHANGELOG.md so that changes can be viewed without leaving the app. See issue [#178](https://github.com/mitre-attack/attack-navigator/issues/178).
 
 ## Fixes
 - Fixed some issues where objects would appear in the wrong domain under specific circumstances. See issue [#308](https://github.com/mitre-attack/attack-navigator/issues/308), [attack-website#310](https://github.com/mitre-attack/attack-website/issues/310).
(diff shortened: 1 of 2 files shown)


nav-app/src/app/data.service.ts
```diff
@@ -327,7 +327,10 @@ export abstract class BaseStix {
         this.id = stixSDO.id;
         this.name = stixSDO.name;
         this.description = stixSDO.description;
-        this.attackID = stixSDO.external_references[0].external_id;
+        if (stixSDO.external_references && stixSDO.external_references[0]) this.attackID = stixSDO.external_references[0].external_id; else {
+          alert('Error: external_id in external_references not found in imported BaseStix object id (' + stixSDO.id + ')');
+          throw console.error('Error: external_id in external_references not found in imported BaseStix object', stixSDO, ' Read more here: https://docs.oasis-open.org/cti/stix/v2.1/os/stix-v2.1-os.html#_72bcfr3t79jx');
```
isaisabel, line 332: Is `console.error()` a valid thing to throw? I would think you'd throw a string or `new Error('desc')` here.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/throw#description

isaisabel: I believe it would in this case throw the value returned by calling console.error, which is presumably null/undefined.

CHANGELOG.md
```diff
@@ -16,6 +16,7 @@ Version 4.4 of the Navigator restores Safari support provided you are using Safa
 - Fixed an issue when loading multiple default layers where subsequent layers would only appear after the user interacted with the first one. See issue [#288](https://github.com/mitre-attack/attack-navigator/issues/288).
 - Fixed an issue with help dropdown appearing above the matrix. See issue [#330](https://github.com/mitre-attack/attack-navigator/issues/330).
 - Updated Safari browser warning to show only for versions 13 and below. See issue [#306](https://github.com/mitre-attack/attack-navigator/issues/306).
+- Added alert to notify user of error in importing a data layer that is missing an ID field. See issue [#339](https://github.com/mitre-attack/attack-navigator/pull/339).
```
isaisabel, line 19: This is incorrect, the user is not importing a layer file but rather loading STIX data. The proper description would be:

> Application now alerts the user when input STIX data contains an object missing an ATT&CK ID instead of failing silently.

nav-app/src/app/data.service.ts
```diff
@@ -327,7 +327,10 @@ export abstract class BaseStix {
         this.id = stixSDO.id;
         this.name = stixSDO.name;
         this.description = stixSDO.description;
-        this.attackID = stixSDO.external_references[0].external_id;
+        if (stixSDO.external_references && stixSDO.external_references[0]) this.attackID = stixSDO.external_references[0].external_id; else {
```
isaisabel, line 330: This won't catch the case where the object has an external_reference but it does not contain the ATT&CK ID. The if statement also needs to check if `external_id` is in the first external reference object.

For reference here's the code we use to deserialize ATT&CK IDs in ATT&CK Workbench: https://github.com/center-for-threat-informed-defense/attack-workbench-frontend/blob/develop/app/src/app/classes/stix/stix-object.ts#L222-L225

isaisabel (changes requested):
Seems like it would work in a few cases but maybe not all. Also has some methodological and documentation improvements to be made.

CHANGELOG.md
```diff
@@ -7,7 +7,7 @@ Version 4.4 of the Navigator restores Safari support provided you are using Safa
 - Improved favicon for standardization with other ATT&CK tools.
 - Show technique even if it's disabled, as long as any of its sub-techniques are enabled. See issue [#298](https://github.com/mitre-attack/attack-navigator/issues/298)
 - Converted help menu to USAGE.md that can be accessed as a raw markdown outside of app. See issue [#179](https://github.com/mitre-attack/attack-navigator/issues/179).
-- Added dialog for CHANGELOG.md so that changes can be viewed without leaving the app. See issue [#178](https://github.com/mitre-attack/attack-navigator/issues/178).
```
isaisabel, line 10: I think you updated the wrong line of the changelog here. It looks like you replaced the changelog line for a different feature entirely instead of changing the prior entry for this feature.

isaisabel (changes requested):
Looks good aside from a documentation snafu.

CHANGELOG.md
iguannalin: Oops, fixed--thanks for catching that!

isaisabel (approved):

PR #342, opened 2021-08-24
Feature/#340 UI scroll refactor

iguannalin:


isaisabel (changes requested):
Summary: fantastic start, all the required changes are there but it needs a bit of cleanup. 

### Toolbar section headers
When the tabs are hidden, the toolbar section headers are as well. This is fine. However, the sleight overlap with the toolbar creates a white pixel within the toolbar, which looks strange. Removing the white border or the overlap should rectify this issue. The white border is for visibility when many layers are open simultaneously.
Screen Shot 2021-08-31 at 2 36 30 PM
Screen Shot 2021-08-31 at 2 36 36 PM

Also interacts strangely with tabs when window is narrow or many tabs are present:
image


### Create layer from other layers compatability
1. Create layer from other layers marks tabs with a variable name for use in the score input. The variable marker is no longer aligned readibly due to the padding removal:
Screen Shot 2021-08-31 at 2 38 52 PM
2. The user can scroll in that interface, hiding the layer variable marker from view when the tabs hide. It may be worth only allowing the tabs to be hidden on scroll when on a layer tab. Alternatively, move the layer variables from the tabs themselves into a list adjacent to the input (would require documentation updates)
3. Image in help page of layer variables probably needs to be recreated with new UI:
Screen Shot 2021-08-31 at 2 44 30 PM

### Bug with changing tabs
If the user scrolls partway down the page and then closes the last tab or switches tabs otherwise, the tab bar may be hidden even though the scroll is now at the top of the screen. This can make the tabs bar inaccessible. To fix this, recompute the amount of hidden tab every time a layer is changed.
image
### Active tab color and style
1. Active tab doesn't need the bottom border color (a separate element of class and type `mat-ink-bar`) since it's already denoted by background color:

image

2. Tab is actually a different color than the toolbar. It should be the same color. This is because `mat-tab-link` has an opacity of 0.6 when not focused. Removing this will improve the UI: 
image

### Invisible help button

For some reason there is an invisible help button to the right of the new tab button.
Screen Shot 2021-08-31 at 2 57 38 PM

Seems to be a white help button that is usually invisible, can be seen when banner is present:

https://user-images.githubusercontent.com/28868888/131561077-21be6152-93c3-46f0-8126-d5b5f9ab6d49.mov



### Closing a tab enables the accessibility scroll arrows
Closing a layer tab for some reason enables the scroll arrows even though there's nothing to scroll to. Opening a new tab removes them.
image

### Banner issues
Banner interacts strangely with tabs, and the help button in the banner is not properly aligned.
image

Banner should scroll out of view alongside tabs, but currently only the tabs area scrolls when the user scrolls.

nav-app/src/app/datatable/data-table.component.ts
```diff
@@ -244,6 +246,25 @@ export class DataTableComponent implements AfterViewInit {
...
+    handleDescriptionDropdown() {
+        this.currentDropdown !== 'description' ? this.currentDropdown = 'description' : this.currentDropdown = '';
+        this.dropdownChange.emit(this.currentDropdown);
+    }
+
+    handleSidenavScroll = () => {
+        const tabHeight = 48;
+        const toolbarHeight = 34;
+        const footerHeight = 33;
+        const tabOffset = Math.floor(Math.min(Math.max(0, this.scrollRef.nativeElement.scrollTop/5), (tabHeight)));
```
isaisabel, line 265: I think this 1:5 ratio of scrolling to tab-hiding is too high. Currently I have to be almost halfway down the page for the tabs to be fully hidden:

image

In my comment on the prior PR I was concerned that 1:2 was even too high, and was suggesting making it 1:1 (removing the `/2` -- now `/5` -- entirely).

isaisabel: With a 1:1 ratio you only need to scroll a little to recover the vertical height of the tabs: 
image

isaisabel (changes requested):
Looks great! I have found a few bugs and have one design concern.

1. the floating help button covers some of the layer controls when scrolled. Maybe the help button shouldn't be floating, and instead scroll out of view when the tabs hide?

   image

2. When the banner is present, the user can create multiple scrollable areas if the viewport is short, which is a bit strange. Depending on mouse position different areas scroll which is very strange to work with:
   image
   image

nav-app/src/app/datatable/data-table.component.ts
iguannalin: I was looking at the [material homepage](https://material.io/) and am considering a different approach, to detect scroll up from scroll down instead. How do you feel about this new behavior? This way, anytime the user scrolls up on the page it would just automatically show the header, instead of calculating how much of the header to show based on their scroll distance.

nav-app/src/app/datatable/data-table.component.ts
```diff
@@ -246,6 +246,8 @@ export class DataTableComponent implements AfterViewInit, OnDestroy {
      */
     ngAfterViewInit(): void {
         // setTimeout(() => this.exportRender(), 500);
+        this.headerHeight = document.querySelector('.header-wrapper')?.offsetHeight;
```
isaisabel, line 249: Seems to have a compile issue:
```
Error: src/app/datatable/data-table.component.ts:249:72 - error TS2339: Property 'offsetHeight' does not exist on type 'Element'.

249         this.headerHeight = document.querySelector('.header-wrapper')?.offsetHeight;
```

PR #347, opened 2021-09-02
Add links/metadata to techniques

### LINKED ISSUE #321: Add ability to assign links to techniques in the context menu
As a user of the ATT&CK Navigator, I would like to be able to assign arbitrary links to particular techniques via the context menu. For example, I would like to be able to assign one or more particular GitHub issues to a technique and access links to those issues via the context menu.

clemiller:
- Adds two new technique controls that allows users to assign links and to add metadata to techniques. Links can be accessed in the context menu; metadata can still be viewed in the tooltip.
- Adds links to the layer level. Links can be added to a layer via the layer information dropdown menu.
- Dividers can be added to link and metadata lists on the technique- and layer-level.
- Upgraded to layer format to v4.3 to support the addition of links to technique objects and layers.
- When creating layers from other layers, users may choose which layer to inherit technique links or metadata from.
- Link and metadata controls are configurable in the "create customized Navigator" interface. 
- Techniques with metadata will have a yellow underline, similar to comments. Techniques with links will have a blue underline. The yellow underline will override the blue underline if both metadata/comments & links exist on a technique.
- The underline color for links can be edited alongside the comment underline color and can be disabled in the configuration file.

Resolves #321


nav-app/src/app/datatable/data-table.component.html
```diff
@@ -496,6 +496,113 @@
...
+                                    <mat-divider></mat-divider>
+                                </div>
+                            </div>
+                        </div>
+                    </div>
+                    <div class="add-more">
+                        <button class="button" (click)="addMetadata()">add {{metadataEditField.length ? 'more' : ''}} metadata</button>
+                    </div>
+                </div>
+             </div>
```
isaisabel, line 603: - Again, should be refactored into its own component.
- Behavior here is similar to the link input -- is it possible to consolidate them into a single component and pass in config to determine the fields to read/write?

nav-app/src/app/datatable/data-table.component.html
```diff
@@ -496,6 +496,113 @@
...
+            <!-- technique metadata -->
+            <div *ngIf="configService.getFeature('metadata')"  class="control-row-item">
+
+                <div class="control-row-button dropdown noselect"
+                     (click)="viewModel.isCurrentlyEditing() && viewModel.metadataMatch ? currentDropdown = currentDropdown !== 'metadata' ? 'metadata' : null : continue"
+                     matTooltipPosition="below"
+                     [matTooltip]="viewModel.metadataMatch ? 'metadata' : 'cannot edit metadata on selected techniques with differing metadata'"
+                     [matTooltipDisabled]="!viewModel.isCurrentlyEditing()"
+                     [style.color]="viewModel.isCurrentlyEditing() && viewModel.metadataMatch ? '#000000' : '#aaaaaa'">
+                     <img [src]="viewModel.isCurrentlyEditing() && viewModel.metadataMatch ? 'assets/icons/ic_metadata_black_24px.svg' : 'assets/icons/ic_metadata_gray_24px.svg'" alt="metadata" />
```
isaisabel, line 562: To work with dark theme this image needs to use material icons instead of importing an image from assets directly.

nav-app/src/app/datatable/data-table.component.html
```diff
@@ -496,6 +496,113 @@
...
+            <!-- links -->
+            <div *ngIf="configService.getFeature('links')"  class="control-row-item">
+
+                <div class="control-row-button dropdown noselect"
+                     (click)="viewModel.isCurrentlyEditing() && viewModel.linksMatch ? currentDropdown = currentDropdown !== 'link' ? 'link' : null : continue"
+                     matTooltipPosition="below"
+                     [matTooltip]="viewModel.linksMatch ? 'link' : 'cannot edit links on selected techniques with differing links'"
+                     [matTooltipDisabled]="!viewModel.isCurrentlyEditing()"
+                     [style.color]="viewModel.isCurrentlyEditing() && viewModel.linksMatch ? '#000000' : '#aaaaaa'">
+                     <img [src]="viewModel.isCurrentlyEditing() && viewModel.linksMatch ? 'assets/icons/ic_link_black_24px.svg' : 'assets/icons/ic_link_gray_24px.svg'" alt="link" />
```
isaisabel, line 509: Again needs dark theme support using material icons features.

nav-app/src/app/datatable/data-table.component.ts
```diff
@@ -375,10 +370,74 @@ export class DataTableComponent implements AfterViewInit, OnDestroy {
...
+    /** Add a new link */
+    addLink(): void {
+        this.linkEditField.push(new Link());
+    }
+
+    /** Remove link */
+    removeLink(i: number): void {
+        this.linkEditField.splice(i, 1);
+        this.viewModel.editSelectedTechniqueValues('links', this.linkEditField);
+    }
```
isaisabel, line 391: Needs to document what the param is -- it's the index, but the doc doesn't say.

nav-app/src/app/datatable/data-table.component.ts
```diff
@@ -375,10 +370,74 @@ export class DataTableComponent implements AfterViewInit, OnDestroy {
...
+    /** Add new metadata */
+    addMetadata(): void {
+        this.metadataEditField.push(new Metadata());
+    }
+
+    /** Remove metadata */
+    removeMetadata(i: number): void {
+        this.metadataEditField.splice(i, 1);
+        this.viewModel.editSelectedTechniqueValues('metadata', this.metadataEditField);
+    }
```
isaisabel, line 408: Again param needs documentation

nav-app/src/app/datatable/data-table.component.ts
```diff
@@ -375,10 +370,74 @@ export class DataTableComponent implements AfterViewInit, OnDestroy {
...
+     *  the current and next indices are not dividers
+     */
+    canAddDivider(i: number, field: string): boolean {
+        let list = this[this.attrToField[field]].filter(value => value.valid());
+        if (list[i] && !list[i].divider) {
+            if (i + 1 >= list.length) return true;
+            if (list[i+1] && !list[i+1].divider) return true;
+        }
+        return false;
+    }
```
isaisabel, line 428: Params need documentation

nav-app/src/app/globals.ts
```diff
@@ -5,5 +5,5 @@
  //
 'use strict';
 
-export const nav_version: string="4.4"
-export const layer_version: string="4.2"
+export const nav_version: string="4.5"
```
isaisabel, line 8: There's going to be a merge conflict here, we now use package.json to track the Navigator version number.

nav-app/src/app/datatable/data-table.component.ts
```diff
@@ -375,10 +370,74 @@ export class DataTableComponent implements AfterViewInit, OnDestroy {
...
+            let divider = new Link();
+            divider.divider = true;
+            list.splice(i+1, 0, divider);
+        } else { // remove divider
+            list.splice(i, 1);
+        }
+        if (field == 'links') this.updateLinks();
+        if (field == 'metadata') this.updateMetadata();
+    }
 }
```
isaisabel, line 443: Params need documentation

nav-app/src/app/matrix/technique-cell/contextmenu/contextmenu.component.html
```diff
@@ -59,5 +59,13 @@
                 {{contextMenuItem.label}}
             </div>
         </div>
+        <div *ngIf="links.length" class="contextMenu-section">
+            <div class="link-container" *ngFor="let link of links">
+                <div *ngIf="!link.divider" class="contextMenu-button" (click)="openLink(link)">
+                    {{link.label}}
+                </div>
+                <div class="divider" *ngIf="link.divider"></div>
```
isaisabel, line 67: This works, but *ngIf also supports else blocks which can help simplify code in some cases. Probably not worth changing here though.

nav-app/src/app/tabs/tabs.component.html
```diff
@@ -251,6 +251,36 @@ <h3>WARNING</h3>
...
+                        <td>
+                            <mat-form-field>
+                                <mat-select placeholder="metadata" [(ngModel)]="metadata" [disabled]="!domain">
+                                    <mat-option [value]="null">none</mat-option>
+                                    <mat-option *ngFor="let vm of getLayers()" [value]=vm>{{vm.name}}</mat-option>
+                                </mat-select>
+                            </mat-form-field>
+                        </td>
+                        <td [ngClass]="{disabled: !domain}">
+                            Choose which layer to import technique metadata from. Leave blank to initialize without metadata.
```
isaisabel, line 280: There's no way to import layer-level metadata, only technique level metadata. We may want to widen the scope of this option to include layer-level metadata.

nav-app/src/app/globals.ts
isaisabel: It doesn't look like the package.json and package-lock.json version numbers have been updated with the new version numbers.

isaisabel (changes requested):
Overall this looks really good from a UX/UI perspective. There are a possible additions to make links and metadata more consistent with existing metadata features, and there are a few small bugs. But overall it looks really good!

- This branch is way behind master and needs to be deconflicted, especially for dark theme support.
- Removing the last link or metadata from a technique will keep dividers even if there's nothing to divide.
- If a divider exists between the first and second metadata/link, and the first metadata/link is removed, a divider will appear in the first position of the interface. Dividers in the first and last position shouldn't be possible, they should only exist between metadata/link
- It is possible to add a divider after the last link/metadata. Dividers should only be possible between items.
- If the metadata/link editing interface is open and the user selects multiple techniques which have non-matching metadata it is still possible to use the interface since it doesn't automatically close, only the button to open it becomes disabled.
   image
- Metadata added at the layer level has a different UI than metadata added at the technique level, and doesn't support dividers. The layer-level metadata interface should be updated to match the technique-level metadata interface.
- Once upon a time the presence of metadata were displayed as yellow underscores like comments. Apparently this feature was dropped somewhere after it was introduced, but should be re-added since it's not possible to see if metadata exists at a glance right now.
    - if a link exists a blue underscore should be used instead. Comment/metadata yellow underscores should override the blue underscore. The blue underscore color should be set in the config file alongside the comment underscore, and should be disable-able in the features structure like comment underscores.

isaisabel:
(sorry, didn't meant to close, I miss-clicked a button)

PR #350, opened 2021-09-10
Layer upgrade feedback

clemiller:
Incorporates the feedback from the v4.4 demo


nav-app/src/app/layer-upgrade/layer-upgrade.component.html
```diff
@@ -13,6 +13,11 @@ <h2>Layer Upgrade</h2>
                 from the knowledge base since ATT&CK {{compareTo.version}} and update 
                 your Navigator layer to reflect these changes.
             </div>
+            <span>
+                <br><a href="https://attack.mitre.org/resources/versions/" target="_blank">
+                    View Versions of ATT&CK
+                </a>
+            </span>
```
isaisabel, line 20: This is sort of a misuse of HTML tags. Theoretically the card content here should be:
- `<div>` containing the `<h1>` title, followed by the versions (as you currently have)
- `<p>` containing the workflow instructions
- `<p>` containing the `<a>` of the version page link.

Generally `<br>` tags shouldn't be used except in specific circumstances (e.g haikus, which are one "block" from a content perspective but contain line breaks for aesthetics). From a formatting perspective using `<p>` blocks is both the most maintainable and also best for accessibility (e.g screen-readers).

isaisabel (changes requested):
Changes look good, but there is a small amount of HTML which should be cleaned up.

isaisabel (approved):

PR #375, opened 2021-11-01
Bugs/#372 dark theme score gradient preset

iguannalin:
Rather than transparent, I changed the color gradient preset to use either the dark theme or light theme cell bg color, as tinycolor does not detect alpha values when calculating the most readable color (i.e. if the bg gradient color for cell A is rgba(255, 255, 255, 0%), and for cell B it is rgba(255, 255, 255, 75%) it will always return black because the most readable color on white is black, even if the 0% transparency in A makes it appear black because of the matrix background in dark theme.


nav-app/src/app/viewmodels.service.ts
```diff
@@ -151,8 +152,8 @@ export class ViewModelsService {
                 // set up gradient according to result range
                 if (score_min != Infinity) result.gradient.minValue = score_min;
                 if (score_max != -Infinity) result.gradient.maxValue = score_max;
-                // if it's a binary range, set to whiteblue gradient
-                if (score_min == 0 && score_max == 1) result.gradient.setGradientPreset("whiteblue");
+                // if it's a binary range, set to transparentblue gradient
+                if (score_min == 0 && score_max == 1) result.gradient.setGradientPreset("transparentblue");
```
isaisabel, line 156: Good catch!

nav-app/src/app/viewmodels.service.ts
```diff
@@ -286,7 +289,16 @@ export class Gradient {
...
-    constructor() { this.setGradientPreset('redgreen'); }
+    constructor() {
+      this.setGradientPreset('redgreen');
+      if (hasCookie("is_user_theme_dark")) this.isDarkTheme = getCookie("is_user_theme_dark") === "true";
+      else this.isDarkTheme = window.matchMedia("(prefers-color-scheme: dark)").matches;
+
+      if (this.isDarkTheme) {
+        this.presets.transparentblue = [new Gcolor("#2e2e3f"), new Gcolor("#66b1ff")];
+        this.presets.transparentred = [new Gcolor("#2e2e3f"), new Gcolor("#ff6666")];
+      }
```
isaisabel, line 300: - If the user creates a layer tab (and thereby calls the constructor for the gradient) on the dark theme, and then later switches to the light theme, the transparentblue and transparentred presets still use the dark theme settings since they don't re-initialize. This essentially locks the layer into the theme it was created under.
- This hardcodes the dark theme background color, which is a variable in the colors SCSS file. If that file changes due to a future revision it won't be updated here -- a small (but probably unavoidable) maintainability concern.

isaisabel (changes requested):
I've opened pr #380 which adds support for alpha colors in gradients. That should allow us to set transparentred and transparentblue endpoints to be actually transparent instead of setting it according to the current theme, which produces bugs (see below).

This fix needs to be documented in the changelog as well.

isaisabel (approved):
Works great!

PR #376, opened 2021-11-03
Fix/#374 dark theme layer upgrade

iguannalin:
Upgrade layer theme window in accordance to the latest dark theme colors

### DIFF
--- a/nav-app/src/styles.scss
+++ b/nav-app/src/styles.scss
@@ -310,7 +310,7 @@ body {
     @include adaptive-color-dark-only("background-color", color(dark-3));
   }
 
-  .mat-expansion-panel-header-title, .mat-icon {
+  .mat-expansion-panel-header-title {
     @include adaptive-color-dark-only("color", on-color(dark));
   }
 
--- a/nav-app/src/styles.scss
+++ b/nav-app/src/styles.scss
@@ -285,6 +285,13 @@ body {
 .mat-drawer-inner-container {
   @include adaptive-color("background-color", color(panel-light), color(dark-2));
 
+  a {
+    @include adaptive-color-dark-only("color", color(dark-link));
+  }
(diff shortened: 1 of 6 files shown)


nav-app/src/styles.scss
```diff
@@ -285,6 +285,13 @@ body {
 .mat-drawer-inner-container {
   @include adaptive-color("background-color", color(panel-light), color(dark-2));
 
+  a {
+    @include adaptive-color-dark-only("color", color(dark-link));
+  }
+  a:visited, a:active {
+    @include adaptive-color-dark-only("color", color(dark-link-active));
+  }
+
```
clemiller, line 294: Nice catch on the links!

clemiller (changes requested):
Overall, this looks great! There are just a couple of small oddities with the UI:

1. There's a bug with the background color of technique cells. Once a changelog cell is rendered in the layer upgrade UI, the background color of the technique cells in the matrix are updated to match the lighter background style of the changelog cells.

Screen Shot 2021-11-04 at 11 08 13 AM

Screen Shot 2021-11-04 at 11 08 31 AM

2. The styles of the "reviewed"/"not reviewed" states of the technique expansion panels need to be updated as well. The "not reviewed" state currently appears like a clickable radio button in dark mode and should be deemphasized to indicate that users can't interact with it. In a "reviewed" state, the checkmark should appear green for consistency between the two themes.

3. Lastly, sort of a trivial detail, there's a discrepancy in the pagination buttons and the mat stepper navigation buttons between light and dark mode. In dark mode, the pagination buttons are given a background color and the navigation buttons appear as a `flat` button rather than `stroked`.
Screen Shot 2021-11-04 at 12 07 23 PM
Screen Shot 2021-11-04 at 12 07 11 PM

PR #377, opened 2021-11-03
Fix loading for multiple layers

### LINKED ISSUE #361: Cannot load multiple default layers from a layerURL
As a user of the application, I want to be able to load multiple layers from a layerURL with the Create a Customized Navigator feature. 

At present this feature seems to exhibit a bug where only the first defined layer loads:
https://mitre-attack.github.io/attack-navigator/#layerURL=https%3A%2F%2Fattack.mitre.org%2Fgroups%2FG0130%2FG0130-enterprise-layer.json&layerURL=https%3A%2F%2Fattack.mitre.org%2Fsoftware%2FS0066%2FS0066-enterprise-layer.json

isaisabel:
Fixes #361. Previously, the await for the first layer to be loaded never resolved, so the application would never attempt to load additional layers. Now the system waits for each layer to finish loading completely (including fetching the STIX data and upgrading layers) before resolving so that the next layer can attempt to load.

Also updates the app version number to v4.5.2.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# v4.5.2 - changes staged on develop
+## Fixes
+- Fixed an issue when loading multiple default layers (specified in the config or a layerURL) which led to only the first layer being loaded. See issue [#361](https://github.com/mitre-attack/attack-navigator/issues/361).
+
 # v4.5.1 - 21 October 2021
 
 ## Fixes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,8 @@
 # v4.5.2 - changes staged on develop
+
+## Improvements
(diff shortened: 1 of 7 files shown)


isaisabel:
Current implementation has a bug with the user upgrading layers that are opened as part of the default set of layers.

```
    "default_layers": {
        "enabled": true,
        "urls": ["https://attack.mitre.org/mitigations/M1015/M1015-enterprise-layer.json", "https://attack.mitre.org/groups/G0130/G0130-enterprise-layer.json", "assets/heatmap_layer.json", "https://attack.mitre.org/mitigations/M1013/M1013-mobile-layer.json"]
    },
```

isaisabel:
User will no longer be prompted to upgrade default layers to the current version of ATT&CK, bypassing the bad UX of being prompted during the initial load.

clemiller (approved):

PR #378, opened 2021-11-05
Fix/#373 system theme color

iguannalin:
System theme color was not changing after setting theme color to dark/light (whichever is opposite of system theme), and switching back to use system settings.

Cookie deletion in HTTPS seems to need the exact same parameters as the ones it was set with, so the same `';path=/;SameSite=Strict'` parameters were added to cookie deletion method.

Tested out cookie removal at: https://iguannalin.github.io/attack-navigator/

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+# Changes staged in develop
+
+## Fixes
+- Fixed inconsistencies within the ATT&CK Navigator dark theme.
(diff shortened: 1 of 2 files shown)


isaisabel (changes requested):
Works fine, and I appreciate your testing on a custom instance. The fix needs to be mentioned in the changelog before this can be merged.

isaisabel (approved):

Issue #359, opened 2021-09-30
Tooltip is too smal for long technique names (or long metadata)


rubinatorz:
Hi there!

For techniques with a long name, like T1048.003, the tooltip when hovering over the technique is too small. It tries to put "Exfiltration Over Unencrypted/Obfuscated Non-C2 Protocol (T1048.003)" in the tooltip but the tooltip is not wide enough:
<img width="735" alt="Screenshot 2021-09-30 at 13 43 39" src="https://user-images.githubusercontent.com/11735227/135449818-01115378-eab2-4386-83fb-fbcd7c0dc5a1.png">

Also when using custom metadata texts, the tooltip can be too small:
<img width="498" alt="Screenshot 2021-09-30 at 13 43 54" src="https://user-images.githubusercontent.com/11735227/135449874-ccb22b58-9dc2-428b-b824-793cfa42a9de.png">

clemiller:
The fix for this is now staged on develop. Closing.

PR #425, opened 2022-04-20
Break-up lines in technique cell tooltip to prevent overflow outside of tooltip box

iguannalin:
Addresses #359


clemiller (approved):
LGTM!

Issue #320, opened 2021-07-13
Scrolling of long comment is not easy to handle


bip2fois:
scrolling through a long comment field with the mouse wheel is a bit difficult.

Scrolling down is manageable in one go.

Back and forth up and down is impossible.

https://user-images.githubusercontent.com/87380145/125478632-a34e6ebb-bd01-4d47-a36b-21069d78c44d.mp4

isaisabel:
The video you attached doesn't seem to work. Are you talking about the comment input field in the toolbar, or the comment display in the tooltip?

UI support for long comments has always been a bit shaky in the Navigator so I can definitely understand if you're having trouble with it.

clemiller:
A new option to `pin/unpin the tooltip` has been added to the context menu (accessed by right-clicking a technique). After clicking on this option, the tooltip will remain open without requiring the mouse to stay hovered over the technique cell and allows the ability to scroll through a long tooltip. This has been merged into `develop` and will be included in the next Navigator release.

PR #426, opened 2022-04-21
Allow user to scroll if comments overflow screen

iguannalin:
Addresses #320

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,12 @@
     The creation of the tag can be disabled with the --no-git-tag-version if desired.
 -->
 
+# v4.6.1 - Changes staged on develop
+
+## Improvements
+
+- Added a feature to pin a technique cell when viewed in the matrix, enabling the user to scroll the details of the technique in the tooltip. See issue [#320](https://github.com/mitre-attack/attack-navigator/issues/320).
+
 # v4.6.0 - 25 April 2022
 
 Adds support for ATT&CK v11.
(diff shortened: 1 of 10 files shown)


clemiller (changes requested):
I really like the idea to have a new option in the context menu! I think the approach needs to be tweaked a bit, since this opens the comment dropdown in the technique controls menu and all other data that can exist in the tooltip (i.e. scores and custom metadata) aren't shown. In cases where there are a lot of metadata values, the problem still persists:

image

One potential solution may be to rename the context menu option to `show tooltip` and edit the `showTooltip()` function in the technique cell component to override the logic and return true if the user has clicked the `show tooltip` option. Then  the override would be removed  when the user clicks outside.  This approach might allow users to easily scroll without the tooltip closing since they wouldn't need to keep their mouse hovered over the cell. This is just one option though and I'm happy to brainstorm some other ideas with you!

CHANGELOG.md
```diff
@@ -16,7 +16,8 @@ Adds support for ATT&CK v11.
...
 
 - Revised the rigid versioning structure in the configuration file which enforced too many constraints on importing custom data. Added new fields `version.version` and `domain.identifier` to improve how Navigator derives domain information. More details are provided in the [integration documentation](https://github.com/center-for-threat-informed-defense/attack-workbench-frontend/blob/develop/docs/integrations.md#attck-navigator-integration) on how to integrate the ATT&CK Navigator with the ATT&CK Workbench. See issue [#370](https://github.com/mitre-attack/attack-navigator/issues/370).
-- Fixed an issue where a long technique name or metadata value would overflow outside of the technique tooltip. See issue [#359](https://github.com/mitre-attack/attack-navigator/issues/359).
+- Fixed an issue where a long technique name or metadata value would overflow outside the technique tooltip. See issue [#359](https://github.com/mitre-attack/attack-navigator/issues/359).
+- Added a feature to pin a technique cell when viewed in the matrix, enabling the user to scroll the details of the technique in the tooltip. See issue [#320](https://github.com/mitre-attack/attack-navigator/issues/320).
```
clemiller, line 20: This should be moved up to a new changelog section for v4.6.1, since v4.6.0 has been released.

nav-app/src/app/matrix/technique-cell/contextmenu/contextmenu.component.html
```diff
@@ -9,6 +9,9 @@
             </div>
         </div>
         <div *ngIf="configService.getFeature('selecting_techniques')" class="contextMenu-section">
+            <div class="contextMenu-button" (click)="pinCell()">
+                pin/unpin this cell
```
clemiller, line 13: I'd recommend renaming this to `pin/unpin tooltip` for clarification and moving it to its own context menu section above the `select/add to selection/remove from selection` options, just to keep the groupings consistent with their action types.

clemiller (commented):
This looks great! It works really well and I think the pinning/scrolling will be very helpful for viewing longer tooltips. There are just a couple of minor documentation/UI things I'd recommend modifying, but after that it's good to merge!

clemiller (approved):
Looks great! Thanks for tackling this issue!

Issue #326, opened 2021-07-29
Uploading a technique with an empty tactic ("") should either ignore the value or reject the layer


isaisabel:
As a user, I want the Navigator to catch improperly formatted layers. 

Currently, uploading a layer with the following technique:
```json
{
    "techniqueID": "T1001",
    "color": "#66ffaa",
    "tactic": "",
    "comment": "example comment",
    "enabled": true,
    "metadata": [],
    "showSubtechniques": false
}
```
Will not show the annotation because the `tactic` is matched to `""`. Since there are no tactics with an empty shortname this will essentially hide the annotation. 

The Navigator should either:
- Treat `""` tactic values as if they are undefined and map the annotation across tactics, or:
- Treat an empty string as an invalid value, and reject the layer from the upload

The current result, while adhering to the layer schema, is very hard to debug since it's easy to assume that an empty tactic string is the same as omitting the tactic field altogether.

PR #429, opened 2022-04-26
Added catch and warning alert if user uploads layer with an empty tactic field

iguannalin:
Addresses #326

### DIFF
--- a/nav-app/src/app/viewmodels.service.ts
+++ b/nav-app/src/app/viewmodels.service.ts
@@ -1652,8 +1652,11 @@ export class TechniqueVM {
         if (techniqueID !== undefined) this.techniqueID = techniqueID;
         else console.error("ERROR: TechniqueID field not present in technique")
 
-        if ("tactic" !== undefined) this.tactic = tactic;
-        else console.error("ERROR: tactic field not present in technique")
+        if (tactic !== undefined && tactic !== "") this.tactic = tactic;
+        else {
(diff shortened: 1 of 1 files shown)


nav-app/src/app/viewmodels.service.ts
```diff
@@ -1652,8 +1652,11 @@ export class TechniqueVM {
         if (techniqueID !== undefined) this.techniqueID = techniqueID;
         else console.error("ERROR: TechniqueID field not present in technique")
 
-        if ("tactic" !== undefined) this.tactic = tactic;
-        else console.error("ERROR: tactic field not present in technique")
+        if (tactic !== undefined && tactic !== "") this.tactic = tactic;
+        else {
+            if ("comment" in obj && obj["comment"] !== undefined) alert(`WARNING: Tactic field for technique ID ${techniqueID} is empty, and will not be displayed in the matrix.`);
```
clemiller, line 1657: Just a couple of suggestions for improvements to clarity and when the warning is displayed:
- The warning alert should be displayed any time the tactic is undefined or an empty string rather than being contingent on whether there is a comment on the technique, since there are other annotations that will be ignored as well
- I'd suggest modifying the warning text a bit- the the technique will still be displayed under the tactic and exist in the matrix. Only the technique annotations will be ignored, so I'd recommend something like:
```
WARNING: The tactic field on the technique <techniqueID> is not defined. Annotations for this technique will not be restored.
```

clemiller (changes requested):

PR #498, opened 2022-09-09
Make json filename lowercase

sergiuser1:


jondricek:
@clemiller - this seems fine, right? If you're good with it, would you mind merging it in? And would you want it in the develop or master branch?

PR #502, opened 2022-09-22
Ignore a matrix's references to tactics that aren't in the source content

ElJocko:
When creating a Matrix object, filters out references to tactics that aren't included in the source content.


clemiller (approved):
Looks good!

Issue #371, opened 2021-10-27
Option to disable the gradient color


rubinatorz:
Is it possible to add a feature to disable the gradient color?

If we now don't provide a gradient object, Navigator uses default colors. And if we provide a gradient object with the color of white, this won't work in the new dark mode.

The reason is that we color the (sub) techniques ourselves (via our DeTT&CT tool) and also want to use the aggregate-sum function, so that the sort is working properly. But we don't want the aggregated techniques to have a color.

PR #538, opened 2023-03-07
Add additional color options

erinehall:
Adds the following three color options to "create a customized navigator"  technique controls.

- Enable/disable displaying colors for aggregate scores
- Enable/disable displaying colors for non-aggregate scores
- Enable/disable displaying manually assigned colors

Resolves [#371](https://github.com/mitre-attack/attack-navigator/issues/371)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,11 @@
     The creation of the tag can be disabled with the --no-git-tag-version if desired.
 -->
 
+# Changes Staged on Develop
+
+## New Features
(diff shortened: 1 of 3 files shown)


clemiller (changes requested):
These changes look great! I have one request to modify the behavior of the "background color" option. Disabling the background color effect works perfectly the way we initially discussed - removing colors for both manually assigned colors _and_ color based on score. However, on more thought, this could also be achieved by disabling all of the new color configuration options, so I think we could change this option to only disable the effect of manually assigned colors. This way users can opt to only view colors based on score.

Aside from that, this just needs an entry in the [changelog](https://github.com/mitre-attack/attack-navigator/blob/develop/CHANGELOG.md)!

Also just wanted to mention that while testing this with the "non-aggregate score color" option disabled, I found a small bug in how aggregate scores are calculated (aggregate scores are calculated on techniques with no sub-techniques). I will open a separate issue to address that.

PR #546, opened 2023-04-03
SVG toolbar dark mode colors

erinehall:
When using Navigator in dark mode, under "render layer to SVG", "image size", the unit suffix for "header height" is black. Additionally, the placeholder text for all the dropdown menus under "render layer to SVG" are black (font, sub-techniques, etc.). This changes their colors to white to match the styling of the rest of the toolbar.


clemiller (changes requested):
This is a much needed improvement, thanks for opening up a PR! There are a couple additional dark mode issues in the SVG interface under the "legend" options:
image

And the drop down arrows in the mat-select fields:
image
image

Would you mind adjusting these as well to include as part of this PR?

clemiller (approved):
LGTM!

PR #572, opened 2023-08-16
New Selection Behavior Option: Select Only Visible Techniques

erinehall:
Added ability to only select visible techniques. Changes include: 

- New selection behavior option "select only visible techniques".
- When selected, hidden techniques are omitted from any selection behavior.


clemiller (changes requested):
This looks great! There are just a couple of minor documentation updates needed and one edge case to be addressed.

clemiller (approved):

Issue #560, opened 2023-06-19
Add "Expand Annotated Sub-Techniques" to layer file format


initstring:
Inside the ATT&CK Navigator UI, there is an option to "expand annotated sub-techniques".

<img width="836" alt="image" src="https://github.com/mitre-attack/attack-navigator/assets/26131150/6d4f910c-3ad4-43ed-aeee-cb94933cc0e9">

It would be nice if we could define this inside a layer file, to ensure this view loads on default.

Perhaps a good spot for this would be in the `layout` section, like this:

```
	"layout": {
		"layout": "flat",
		"expandAnnotatedSubtechniques": true
	},
```

We use Navigator inside a CI pipeline, which automatically generates layer files and builds a static website for us to compare individual and combined red team operation output. For us, it is a dashboard we can visit to view our progress over time. Having this view defined automatically would ensure everyone gets a consistent experience, and having these expanded by default ensures people's eyes are drawn to the specific techniques we have annotated.

Thank you for all your hard work on a great tool!

clemiller:
Changes for this issue have been staged on `develop` and will be included in the next release.

An optional property called `expandedSubtechniques` will be added to the `layout` section in the Layer file and can be set to `all`, `annotated`, or `none`. This will expand all subtechniques, expand only annotated sub-techniques, or collapse all sub-techniques respectively. Thank you for your suggestion!

initstring:
Thanks so much for your work on this, @clemiller!

I know this issue is closed, but I was hoping to also draw attention to [this other issue](https://github.com/mitre-attack/attack-navigator/issues/123#issuecomment-1596274133), as it is related. Perhaps it could be addressed in the same way?

Thanks again for your hard work on this great tool.

clemiller:
Hi @initstring, I have added #123 to our next milestone and will have our team start looking into it following the upcoming release.

PR #578, opened 2023-09-14
Expand annotated subtechniques

adpare:
This pull request address #560 

Users should now be able to configure how to display sub techniques in the layer file - annotated, all, or none

Includes the metadata_underline change

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,7 +25,8 @@
 
 ## Layer File Format Changes
 
-Layer file format updated to version 4.5. This update adds support for selecting only visible techniques. The `selectVisibleTechniques` field specifies whether or not hidden techniques will be included in the different select behaviors. See [layers/LAYERFORMATv4_5.md](layers/LAYERFORMATv4_5.md) for the full specification.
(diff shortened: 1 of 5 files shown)


clemiller (changes requested):
Overall, this looks great - the majority of edge cases are working as expected, there's just one more edge case that needs to be addressed, as well as a couple of updates to the changelog:
- There is one case where loading an existing layer doesn't display as expected. The existing configuration for any techniques in the layer with `"showSubtechniques": true` is not honored when `expandedSubtechniques` is set to "none".  Since "none" is the default value, any layer prior to this feature will never show any sub-techniques. The `"showSubtechniques": true` in the layer should take precedence over the `expandedSubtechniques` settings.
- Check out the reported SonarCloud issues to see if there are any improvements that could be made, looks like there's one code smell that could easily be fixed
- A description of the changes to the layer format should be added to the changelog under the "Layer File Format Changes" section
- Needs an entry under "New Features" in the changelog referencing issue 560 and the Layer File Format Changes section

clemiller (approved):
LGTM!

Issue #484, opened 2022-07-28
Add option for choosing Font Size when Exporting to SVG


jondricek:
Related to #221 and #394.

## Problem
When exporting to SVG from Navigator, the font size is consistent ever since @lilyjw's pull request in #394. This is better than the dynamic font size approach from before, but it can be pretty small to read.

## Request
It would be nice to be able to set a font size when exporting, so not only would the font be consistent, but configurable to the end user.

PR #584, opened 2023-10-02
Option to Change SVG Font Size

erinehall:
Added the ability for users to set the font size of the SVG export. Changes include:

- New "auto-fit font size" button, which calculates the most visually appealing font size
- New font size input, which allows users to manually change the SVG font size
- Warning message if the font size is increased such that the text overflows outside of any of the cells

Resolves https://github.com/mitre-attack/attack-navigator/issues/484


clemiller (approved):
LGTM!

Issue #508, opened 2022-10-07
Firefox screenshot scrolling not working because of .matrices style


wasserman:
I'm trying to do a scrolling screenshot with Firefox, which generally works.  I see you have Excel and SVG exports, but I prefer having the appearance that matches what I see in in the navigator.  This is on version v4.6.6 from https://mitre-attack.github.io/attack-navigator/#comment_underline=false.

I noticed that the inner scroller can be removed by removing the style height definition in the html.

It shows up like this:
```
<div class="matrices" style="height: calc(-114px + 100vh);">
```

If I change it via the Inspector in Web Dev Tools to this then I can take a Firefox screenshot properly via the `Save full page` option.
```
<div class="matrices" style="">
```  

Could some option or setting be added to allow the removal of this scrollbar natively?  I don't know the purpose of it, but I assume it is there for a reason. 

Thanks!

wasserman:
Reference for Firefox screenshots: https://support.mozilla.org/en-US/kb/take-screenshots-firefox

clemiller:
Hi @wasserman,

The height of the matrices is defined to allow the toolbar to remain in place via "sticky" positioning while scrolling to the bottom of the matrix. This was implemented as part of the tabs refactor in [#340](https://github.com/mitre-attack/attack-navigator/issues/340). 

![image](https://user-images.githubusercontent.com/48832936/194554340-11a3e7fe-9a50-42d5-a636-97ffb6571d4c.png)

This issue looks potentially related to [#358](https://github.com/mitre-attack/attack-navigator/issues/358), though I'm unsure if it's the toolbar that's causing issues with your screenshots. At some point there was a configuration setting added to remove the scrolling causing issues with full-page screenshots, but it appears it has since been removed. Keeping this issue open to either re-implement the configuration setting or see if we can find a better approach.

wasserman:
Thanks @clemiller.  I honestly only even tried to do the screenshot because the print view was a bit wild.  Maybe some custom css for print media would be a better improvement or something else to consider?

https://developer.mozilla.org/en-US/docs/Web/CSS/@media

clemiller:
Changes for this issue have been staged on `develop` and will be included in the upcoming Navigator release. This adds both custom css for improved printing views and a button in the toolbar to enable/disable the sticky toolbar. Thank you for your suggestions!

PR #589, opened 2023-10-09
508 full page screenshots

adpare:
This pull request addresses #508.

- Added functionality to display matrices accurately in print view.
- Added functionality to enable or disable the stickiness of the toolbar.


nav-app/src/assets/config.json
```diff
@@ -244,6 +244,9 @@
             {"name": "link_underline", "enabled": true, "description": "Disable to remove the hyperlink underline effect on techniques."},
             {"name": "metadata", "enabled": true, "description": "Disable to remove the ability to add metadata to techniques."},
             {"name": "clear_annotations", "enabled": true, "description": "Disable to remove the button to clear all annotations on the selected techniques."}
+        ]},
+        {"name": "toolbar_controls", "enabled": true, "description": "Disable to disable all subfeatures", "subfeatures": [
+            {"name": "sticky_toolbar", "enabled": true, "description": "Disable to remove the sticky toolbar."}
```
clemiller, line 249: I'd recommend updating the description of "sticky_toolbar" to something like: "Disable to remove the ability to enable/disable the sticky toolbar". The current description implies that the sticky toolbar will be disabled s.t. it will not remain at the top of the screen on scroll, if a user changes this to `"enabled": false`. Disabling this feature will only remove the icon from the toolbar, but the sticky toolbar will still be enabled by default.

clemiller (changes requested):

Issue #404, opened 2022-01-28
Freeze the Tactic Row while Scrolling


jondricek:
When using the Navigator, it would be useful to be able to scroll down with the Tactics header frozen to the top. Not sure how difficult this would be to implement, but definitely a feature worth investigating.

There is a sister issue for this for the main website here: https://github.com/mitre-attack/attack-website/issues/350

PR #592, opened 2023-10-17
Freeze the tactic row

adpare:
This pull request addresses #404.

Added functionality to scroll down with the Tactics header frozen to the top.


clemiller (changes requested):
There are a couple of bugs that need to be addressed before merging:

- The sticky tactic header is only set for the side layout, it should follow the same behavior for the flat matrix layout as well
- The tactic row collapses oddly when the ATT&CK IDs are enabled. The tactic name is either fully or partially hidden on scroll, where it would be expected for the name to still be shown:
image
- The tactic row background color does not change when enabled/set. The text color still changes accurately based on the expected background color, but the background color is not updated:

https://github.com/mitre-attack/attack-navigator/assets/48832936/1e3eccec-a3ba-41b2-aa82-3c6de67910cb

Issue #505, opened 2022-09-28
Merging of multiple bundles into a single matrix


jfcantu:
Hi folks,

I'm not sure if I'm doing something wrong here, or if this isn't supported, but here's my use case:

According to the instructions [here](https://github.com/mitre-attack/attack-navigator/#loading-content-from-local-files), a "domain" (as defined in `config.json`) can be configured to load multiple STIX bundles.

It seems that if you load multiple STIX bundles in this way, objects in one bundle can't be referenced from another bundle. For example, if I load the standard Enterprise ATT&CK bundle, I can't add onto that matrix by loading a second bundle containing Techniques which reference Tactics in the ATT&CK bundle.

This would be helpful for a couple reasons - for one, I would like to be able to extend the ATT&CK framework without having to maintain my own fork of the ATT&CK STIX bundle. Secondly, STIX bundles are already unwieldy enough - it would be very helpful to be able to split objects across multiple files for ease of maintainability.

clemiller:
Hi @jfcantu,

Thanks for opening this issue! This appears to be caused by a bug in the bundle parsing.

For additional context- as per [#183](https://github.com/mitre-attack/attack-navigator/issues/183), the Navigator should allow users to define multiple bundles to be loaded into a single Navigator instance, such as if domain data is split across multiple bundles. We previously allowed displaying multiple domains in a single layer, but due to issues with layer-layer operations and custom datasets, we removed that ability by filtering out objects that are not within the specified domain ([#308](https://github.com/mitre-attack/attack-navigator/issues/308)).

Users should still be able to specify multiple bundles that are within the same domain, however, each bundle is parsed and built separately with the matrix and tactics provided. Because of this, all subsequent bundles that either do not include their own matrix or contain a matrix with the same STIX ID as a previous bundle are not built correctly and are not displayed in the Navigator view. The bundle parsing should have the ability build multiple bundles into a single matrix view.

clemiller:
Fixes for this issue have been staged on `develop` and will be included in the next Navigator release.

PR #613, opened 2024-02-07
505 merging of multiple bundles into a single matrix

adpare:
This PR addresses issue #505.


nav-app/src/app/services/data.service.ts
```diff
@@ -220,6 +229,19 @@ export class DataService {
...
+        //create matrices, which also creates tactics and filters techniques
+        for (let i = 0; i < matricesList.length; i++) {
+            let techniquesList = [];
+            if (matricesList[i].x_mitre_deprecated) {
+                continue;
+            }
+            for (let j = 0; j < domain.techniques.length; j++) {
+                if(domain.techniques[j].get_technique_domain(techniqueSDOslist[j]) == matricesList[i].external_references[0].external_id) {
+                    techniquesList.push(domain.techniques[j]);
+                }
```
clemiller, line 241: This logic should be updated to check if the matrix ID matches any of the domains listed in the technique's `x_mitre_domains` field. See comment above regarding only checking the first domain in the list of `x_mitre_domains`.

clemiller (changes requested):
This looks great. The functionality to view multiple bundles/matrices is working. I have one comment on the logic for checking if a technique is part of a matrix/domain.

clemiller (commented):
Looks great! Thanks for making these updates. Please take a look at the two issues reported by SonarCloud and see if they can be addressed. Once that's done, we can merge this in!

Issue #608, opened 2024-01-25
Feature request: filter on ATT&CK data sources


RobbeVandenDaele:
Hi all, 

I had an idea which I wanted to share with you. It think it would be a great feature if we can filter the navigator based on the ATT&CK data sources (https://attack.mitre.org/datasources/). This would help a lot in plotting detections on the navigator related to a specific data source, since this removes the clutter of techniques not relevant to a specific data source. I feel like the filters for the platforms are not always sufficient enough.

Kind regards
Robbe

clemiller:
Hi @RobbeVandenDaele,

Thank you for opening up this suggestion! I agree that the filtering options in Navigator could be improved and this is an interesting idea. I am leaving this issue open for our team to consider and will mark it as "seeking feedback" in case others in the community have additional thoughts they would like to share.

In the meantime, there are a number of steps you can take to achieve similar functionality. The process for hiding techniques is somewhat involved (related to the improvements requested in #571), but I'm sharing it here in case it is useful to you:
1. Open the `search & multiselect` sidebar
2. Select techniques based on a specific Data Source from the `Data Sources` panel
3. Right click a technique in the matrix view and choose "invert selection"
4. Select "toggle state" under technique controls in the toolbar
5. Click "show/hide disabled" under layer controls in the toolbar

vynttran:
Hi @RobbeVandenDaele and @clemiller, 

This indeed seems like a useful feature to add! I think we can implement this by adding a ‘data sources’ section in the filters menu (relatively similar to the current ‘platforms’ section, but the toggles in the section will be based on the different data sources). 

That said, would it be possible for me to work on this issue? On my team are 3 other friends, @csuraparaju, @tarunBandi-ONE, and @ytw-wyt, who are also excited to help out with implementing this.

clemiller:
Data sources were deprecated in ATT&CK v18 (October 2025) and are no longer part of the current model, so we won’t be implementing data source–related features. We’ll close this issue, but we appreciate your interest in improving the Navigator.

PR #633, opened 2024-04-23
Feature Addition: Filter by Data Sources 

vynttran:
- Resolves feature request in #608
- Description: With this new addition, the filter menu dropdown would have not just the filter by platforms section, but also a filter by data sources section. 
- Implementation: As a search based on data sources feature already exists in the current version of the Navigator, we built this addition based off of that, which should ensure the technique-to-source mappings are accurate, and so is this added filtering functionality. Below is a screenshot of what the addition would look like: 
Screenshot 2024-04-23 at 2 22 55 PM

### DIFF
--- a/nav-app/angular.json
+++ b/nav-app/angular.json
@@ -142,5 +142,8 @@
         "@schematics/angular:directive": {
             "prefix": "app"
         }
+    },
+    "cli": {
+      "analytics": false
     }
 }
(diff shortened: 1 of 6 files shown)


clemiller:
Hi all, please ping me when this PR is ready for review and I will take a look. Thanks!

GovindarajanL (approved):

clemiller:
Thanks for opening this draft and exploring a data source filter for the Navigator. Since the v18 release (October 2025), data sources have been deprecated and are no longer part of the current model, so we’re not adding new data source-based features. With that in mind, we’ll be closing this PR. Thanks again for the work you put into it.

PR #640, opened 2024-05-16
 fix: nav-app/package.json and Dockerfile to reduce vulnerabilities

b34c0n5:


clemiller:
Hi @b34c0n5,

Thanks for your interest in improving the Navigator. Unfortunately, we cannot merge this PR at this time. The changes proposed include bumping Node.js to a version that our application currently does not support. Additionally, the upgrade to `ngx-markdown` v17 is not compatible with the version of Angular that Navigator uses (v14). We are in the process of upgrading to Angular v17, but do not have a timeline on when that will be completed/released. We are always open to improvements and upgrades, but they need to be compatible with our current setup. Thanks for your understanding.

PR #666, opened 2024-11-06
Update Dockerfile for production

pokorny-martin:
Current Dockerfile is not suited for production use.  
We've created one using multistage build to build the dist before starting the server and serving the files using unprivileged nginx.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,27 +1,11 @@
-# Build stage
-
-FROM node:18
-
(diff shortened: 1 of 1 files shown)


clemiller:
Thank you for proposing an improved Docker setup. We’ve recently updated the Dockerfile to a more production-oriented configuration that aligns with Angular 19, so this PR is now out of date. I’ll close it here, but we appreciate you taking the time to contribute.

PR #687, opened 2025-07-18
Update gradient.ts

CVE-1990-0001:
Prune value to the maximum possible index of the gradientRGB array.

This is solving the following bug with a Math.round(r) value = 100 creating an array out of bounds exception:

image

image


erinehall:
Looks good to me!

PR #826, opened 2026-09-09
ci: run release workflows from version tags

scotluns:
## Summary

Update release-side GitHub Actions workflows so public release artifacts are produced from explicit stable version tags instead of ordinary branch pushes.

This changes:

- GitHub Pages deploys to run on stable `vX.Y.Z` tag pushes instead of `master` pushes
- GitHub Release creation to run only on stable `vX.Y.Z` tag pushes
- Docker image publishing to run only on stable `vX.Y.Z` tag pushes

Docker publishing still produces the release semver tag, `latest`, and `sha-*` tags. With this change, `latest` represents the latest tagged release image rather than the latest `master` branch push.

## Validation

Validated the trigger behavior on a fork before opening this PR:

- Branch pushes did not trigger release-side workflows
- A disposable `v*.*.*` tag triggered GitHub Pages, Create Release, and Docker Publish workflows
- Docker Publish was tested with a dry-run echo step on the fork to confirm trigger behavior without publishing an image

## Notes

- This PR intentionally does not include the existing `develop` CI quality-gate changes.
- GitHub Actions tag filters use glob matching, so each release workflow now validates `GITHUB_REF_NAME` before creating release artifacts.

### DIFF
--- a/.github/workflows/create-release.yml
+++ b/.github/workflows/create-release.yml
@@ -3,7 +3,7 @@ name: Create Release
 on:
   push:
     tags:
-      - '*'
+      - 'v*.*.*'
 
 jobs:
   create_github_release:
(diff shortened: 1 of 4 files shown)


jondricek (approved):
lgtm

PR #830, opened 2026-09-14
test: establish E2E foundation with initial smoke coverage

scotluns:
## Why

Establish a foundation for E2E tests that gives contributors confidence that changes preserve Navigator's core workflows. Begin with smoke coverage and a minimal happy path, then add scenarios within the same test setup as coverage grows.

## What Changed

- Adds a Chromium Playwright suite under `nav-app/e2e`, run with `npm run test:e2e`. Tests tagged `@smoke` can be selected with `--grep @smoke`; the README documents the convention and shared setup.
- Starts with one happy-path smoke test: create an Enterprise layer, select and score a technique, rename the layer, and validate the downloaded JSON. It uses existing synthetic STIX fixtures and existing UI selectors, without application changes.
- Runs the suite against the production build in GitHub Actions, with one worker and no retries. Artifacts go under `tmp/playwright/test-results`; CI uploads failure results and traces for seven days.
- Documents local and optional Docker execution; the container runs tests as a non-root user. Protractor remains untouched for removal in a separate change.

## How To Verify

- Follow `nav-app/e2e/README.md` to build and run locally. The current tagged smoke test passes and writes its export to the configured artifact directory.
- The current revision [passed GitHub CI](https://github.com/mitre-attack/attack-navigator/actions/runs/34878339958) and SonarCloud checks. A [deliberately incorrect score assertion](https://github.com/mitre-attack/attack-navigator/actions/runs/34864439272) confirmed CI fails and uploads the exported JSON and a readable trace; that assertion was restored.
- The current Docker image builds and the smoke test passes on Linux arm64 as a non-root user, using the documented run command without platform or shared-memory overrides.

## Known Gaps

Initial coverage uses synthetic data and checks interaction and export behavior. Full ATT&CK data, layer comparison, import, visual rendering, and other browsers remain future coverage.


nav-app/e2e/layer.spec.ts
```diff
@@ -0,0 +1,54 @@
...
+    await page.route('**/assets/config.json', route => route.fulfill({
+        json: { ...config, collection_index_url: '', versions: configData },
+    }));
+    await page.route(configData.entries[0].domains[0].data[0], route => route.fulfill({
+        json: {
+            type: 'bundle',
+            id: 'bundle-smoke',
+            objects: [matrixSDO, TA0000, { ...T0001, name: 'Smoke technique' }, { ...T0003, name: 'Untouched technique' }],
+        },
+    }));
```
jondricek, line 20: This looks good, but to me it is highlighting that our mock data that it uses here hasn't been updated in 2+ years and probably longer than that, since https://github.com/mitre-attack/attack-navigator/blob/master/nav-app/src/tests/utils/mock-data.ts has ATT&CK v13 in it which is from April 2023.

jondricek (approved):
Overall I approve this pull request, but i'd like you to open a new GitHub issue to update our approach to using mock data for tests in ATT&CK Navigator. We should probably continue to test with various versions of ATT&CK, so we shouldn't completely discount old stuff, but we should also test with current data too, whether mock data or real ATT&CK releases

